PR #3, opened 2021-05-27
Structs

rahulanand16nov:
Added common structs so Lahiru and I can start working on filter and singleton. Let me know what do you think.


threescale/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+pub mod cache;
+
+// Request data recieved from previous filters
+#[allow(dead_code)]
+pub struct ThreescaleData<'b> {
+    app_id: String,
+    service_id: String,
+    metrics: RefCell<HashMap<&'b str,u32>>,
+}
+
```
NomadXD, line 12: EOF new line

NomadXD (approved):
LGTM. Approved

rahulanand16nov:
cc @unleashed

threescale/src/cache.rs
```diff
@@ -0,0 +1,38 @@
...
+  max_value: u32,
+}
+
+#[allow(dead_code)]
+pub struct Application<'a> {
+  app_id: String,
+  service_id: String,
+  timestamp: Duration,
+  local_state: RefCell<HashMap<&'a str,UsageReport>>,
+  remote_state: RefCell<HashMap<&'a str,UsageReport>>,
```
unleashed, line 36: This reminds me a lot about the Go cache, which makes this distinction - I'm not sure the approach in that cache using these two representations of state is the best start here - I think it is complex and a simpler way to deal with this might exist, perhaps we just need to think about the cache workflow and see whether we really do need these two maps.

rahulanand16nov: Just changed to one map that counts left hits instead of tracking remote and local states individually.

PR #4, opened 2021-05-28
Flow definitions

rahulanand16nov:
Creating PR to get feedback on any case I am missing and provide @NomadXD with sample functions to call from the singleton.
LMK.
cc @unleashed


cache-filter/src/filter.rs
```diff
@@ -69,7 +81,49 @@ struct CacheFilter {
...
+impl HttpContext for CacheFilter {
+    fn on_http_request_headers(&mut self, context_id: usize) -> Action {
+        let current_time = self.get_current_time();
+        let queue_id = self.resolve_shared_queue(VM_ID, QUEUE_NAME);
+        let request_data = match get_request_data() {
+            Some(data) => data,
+            None => {
+                info!("ctxt {}: Releveant request data not recieved from previous filter", context_id);
+                // Send back local response for not providing relevant request data
+                self.send_http_response(404,vec![],None);
```
NomadXD, line 93: I think sending a 401 is more suitable here because otherwise it will mean that the resource user requested doesn't exist on the upstream.

rahulanand16nov: Updated. ccing again @unleashed because the previous one didn't update the reviewer's list.

NomadXD:
LGTM. Approved

PR #5, opened 2021-05-30
Refactoring, threescale-utils, cache hit flow implemented

rahulanand16nov:
Touched a lot of files so, please have a close look and let me know if I made any errors.


cache-filter/src/filter/http.rs
```diff
@@ -0,0 +1,125 @@
...
+            }*/
+    
+            // If any metric is rate-limited then whole request is restricted
+            if app.local_state.borrow().get(metric).unwrap().left_hits - hits < 0 {
+                true;
+            }
+        }
+
+        if !set_application_to_cache(self.cache_key.as_str(), app, true, None) {
+            self.to_report_singleton = true;
```
NomadXD, line 121: This should be `false` since `set_application_to_cache` returns `false` when application is written to cache. Now the value of `to_report_singleton` is always `true` because it's initialized as `true` and here it's updated as `true`. I think it's better to change the `set_application_to_cache` logic so that it will return `true` when set application succeeds. Now with the inverted logic it's confusing and I think that's why you have mistakenly set up true here.

cache-filter/src/filter/root.rs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+    }
+
+    fn create_http_context(&self, context: u32) -> Option<Box<dyn HttpContext>> {
+        Some(Box::new(CacheFilter {
+            context_id: context,
+            config: self.config.clone(),
+            to_report_singleton: true,
+            cache_key: String::new(),
```
NomadXD, line 60: Better to have this as a string literal in a function scope where it's needed. `String` will be allocated on the heap and since the `HTTPContext` gets created per every http request, we'll have heap allocations equal to the number of requests. String literals are in the stack and gets invalidated when it goes out of scope, it's better to have this as a string literal.

cache-filter/src/filter/http.rs
rahulanand16nov: Good catch! I'll change the return pattern and set the default value of `to_report_singleton` to `false`.

cache-filter/src/filter/root.rs
rahulanand16nov: We don't have a strict memory restriction but rather a time restriction so, no need to worry about heap/stack allocation. I chose `String` because it's a matter of who owns that memory. In the case of `String`, `CacheFilter` owns that matter and in the case of string literal, that memory lies outside the `CacheFilter`.

threescale/src/utils.rs
```diff
@@ -1,13 +1,58 @@
-use crate::structs::Application;
+use crate::structs::{ Application, PeriodWindow };
+use proxy_wasm::hostcalls::{ set_shared_data, get_shared_data};
+use log::info;
+use std::time::{ SystemTime, Duration };
 
 // Returns Application from shared data with CAS integer
 pub fn get_application_from_cache(key: &str) -> Option<(Application,u32)> {
-    None
+    match get_shared_data(&key).unwrap()
```
unleashed, line 8: I'd prefer that when `get_shared_data()` fails we not only panic but also we have an explicit path in the code panicking, such as:
```rust
match get_shared_data(&key) {
    Ok(Some(bytes), cas) => { ... // deal with happy path },
    Ok(None, cas) => { ... // handle no data for this key - perhaps panic for the time being },
    Err(status) => { ... // handle not found for this key - perhaps panic for the time being },
}
```

In general most, but not all, `unwrap()` calls mean that there is some match arm or else clause that needs to be handled (ie. we want to avoid panics), so having them explicitly show up in the code, even if just panicking, helps people having them fixed earlier than otherwise.

threescale/src/utils.rs
```diff
@@ -1,13 +1,58 @@
...
+    {
+        (bytes,cas) => {
+            match bytes {
+                Some(data) => Some((bincode::deserialize::<Application>(&data).unwrap(),cas.unwrap())),
```
unleashed, line 12: Note that it is possible that `cas` might be `None` for this case, and `unwrap()` would then panic. If you prefer to deal with `cas` elsewhere, you might want to change the signature to return `Option<(Application, Option<u32>)>`.

threescale/src/utils.rs
```diff
@@ -1,13 +1,58 @@
...
+                Some(_c) => cas_o,
+                None => None,
+            }
+        }
+    }
+}
+
+// Returns false on set failure
+pub fn set_application_to_cache(key: &str, app: &Application, overwrite: bool, num_tries: Option<u32>) -> bool {
+    let mut cas = Some(0);
```
unleashed, line 33: Just passing in `None` is symmetric with `get_shared_data` (returns `None` if it is actually `0`) and behaves as if it was `Some(0)`.

threescale/src/utils.rs
```diff
@@ -1,13 +1,58 @@
...
+            Ok(()) => return true,
+            Err(e) => info!("Try {}: Set operation failed for key: {} due to: {:?}", num_try, key, e),
+        }
+        cas = get_cas_from_cache(key);
+    }
+    false
+}
+
+pub fn get_next_period_window(old_window: &PeriodWindow, current_time: &SystemTime) -> PeriodWindow {
+    // TODO: How to calculate next window?
```
unleashed, line 52: Periods are absolute (natural calendar periods) based on UTC. So a day would run from 00:00:00 UTC to 23:59:59.999 UTC, an hour would run from 10:00:00 UTC to 10:59:59.999 UTC, etc. Hopefully we can come up with something where we can rely on having frequent flushing of data to 3scale to avoid having to care about aggregating lots of periods. The information about periods comes from 3scale when you call authorize to get the status of limited metrics (the usage reports): all limits will have attached period type, start and end fields.

cache-filter/src/filter/root.rs
unleashed: Note that string literals are not in the stack, they are statically allocated when the binary is loaded and are immutable - if you need to compute the `cache_key` at runtime then you cannot use a string literal. You actually compute it at runtime with `self.cache_key = format!("{}_{}",request_data.app_id,request_data.service_id);`, so it's going to be an allocation in the heap.

That said, while there are some (convoluted) ways to avoid heap allocation it is not an issue until the functionality is complete and correct, and at that point we should still assess whether it could result in a big enough impact that merits going for it.

unleashed (commented):
A couple of suggestions to use Rust tooling:

1. Ensure you run `cargo fmt` on the project to have idiomatic indentation and whitespace.
2. Ensure you also run `cargo clippy` on the project to improve on idioms that help both the code and reviewers better understand it. You can add `#![deny(clippy::all, clippy::cargo)]` to `src/lib.rs` to get started with useful suggestions.
3. Make it automatic via continuous integration - there are some Github Actions you could use which are mostly copy&paste to the project. Ping me if you need help, but you could take it from the other "auth" filter.

cache-filter/src/filter/root.rs
NomadXD: Sorry my bad. Only the reference is in the stack and the memory is allocated from the heap.

threescale/src/utils.rs
rahulanand16nov: Thanks for the great feedback! I'll keep them in mind and add the workflow to automate as mentioned by you.

threescale/src/utils.rs
rahulanand16nov: Fixed.

threescale/src/utils.rs
rahulanand16nov: Made the suggest changes.

NomadXD (approved):
LGTM, approved

PR #6, opened 2021-05-31
Dev/singleton service

NomadXD:
- proxy module added to threescale package for proxy related common functionalities
- delta store implemented
- LICENSE and NOTICE added


singleton-service/src/service/proxy.rs
```diff
@@ -0,0 +1,155 @@
...
+    delta_store: DeltaStore,
+}
+
+impl RootContext for SingletonService {
+    /// Message queue will get registered when on_vm_start callback gets called.
+    fn on_vm_start(&mut self, _vm_configuration_size: usize) -> bool {
+        if let Ok(q_id) = register_shared_queue(QUEUE_NAME) {
+            self.queue_id = Some(q_id);
+        }
+        // TODO : handle MQ failure, change info to trace after dev
```
rahulanand16nov, line 50: Failing to register proxy will cause the proxy to crash so you don't have to worry about this.

singleton-service/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 mod configuration;
 mod service;
+mod deltas;
```
unleashed: Nit: lacks newline before EOF (same with deltas.rs).

singleton-service/src/service/proxy.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    /// Configuration passed by envoy.yaml will get deserialized to ServiceConfig. If there's an issue with the
+    /// passed configuration, default configuration will be used.
+    fn on_configure(&mut self, _config_size: usize) -> bool {
+        // Check for the configuration passed by envoy.yaml
+        let configuration: Vec<u8> = match self.get_configuration() {
+            Some(c) => c,
+            None => {
+                warn!("Configuration missing. Please check the envoy.yaml file for filter configuration");
+                return false;
```
rahulanand16nov, line 66: We already have a default config, so we don't have to cause a panic if the configuration is missing. Please set this to 'true`.

singleton-service/src/service/proxy.rs
unleashed: The proxy should never crash - the host implementation might panic, which is a well known bad practice, and hopefully we'll avoid that in the future and provide a status from the SDK that can be helpful to log and diagnose the issue. IOW, just don't depend on code that you don't control (the SDK) but instead depend on the full signature (ie. it can be an `Err(...)`) - this way we can have the SDK fixed without having to hunt for implicit dependencies in our code.

NOTICE
```diff
@@ -0,0 +1,14 @@
+   GSoC '21 Proxy-WASM filters - 3scale labs
+   Copyright (c) 2020, 2021 Red Hat, Inc.
```
unleashed, line 2: There's no copyright attribution required for 3scale/Red Hat unless you sign such a thing - in other words, you can (and should) use your own copyright :)

singleton-service/src/service/proxy.rs
```diff
@@ -0,0 +1,155 @@
...
+                );
+                match queue_entry {
+                    Some(threescale) => {
+                        // TODO: Hanlde deserialization safely.
+                        let threescale: ThreescaleData = bincode::deserialize(&threescale).unwrap();
+                        // TODO: If update required, then update, Handle update failure
+                        self.delta_store.update_delta_store(&threescale);
+                    }
+                    None => {
+                        info!("No application found from the message queue entry")
```
rahulanand16nov, line 107: We are now consuming a Message struct(which contains the flag), not a Threescale struct. It's present inside `threescale::structs`.

singleton-service/src/service/proxy.rs
```diff
@@ -0,0 +1,155 @@
+pub use crate::configuration::service::ServiceConfig;
+pub use crate::service::deltas::DeltaStore;
+use log::{debug, info, warn};
+use proxy_wasm::{
+    hostcalls::{dequeue_shared_queue, register_shared_queue},
```
rahulanand16nov, line 5: Just a small nit: You can call host functions using `self.` inside implementation for any Context (Root or Http). You can remove these imports and use `self.host_call`.

rahulanand16nov (changes requested):
Just apply mentioned changes and, everything looks great!

singleton-service/src/service/proxy.rs
rahulanand16nov: I'll write these design problems from the SDK side inside the design doc for later use.

singleton-service/src/service/proxy.rs
NomadXD: fixed

singleton-service/src/service/proxy.rs
NomadXD: I was using the default config for the scenario of config provided and parsing fails. But yeah even for the scenario where the user hasn't provided a config, we can go with the default.

singleton-service/src/service/proxy.rs
NomadXD: yeah but there's a slight difference in the APIs. The APIs in the hostcalls are wrapped with `Result`

NOTICE
NomadXD: fixed

rahulanand16nov (approved):

PR #9, opened 2021-06-04
Report request to 3scale SM API implemented

NomadXD:
Report request to 3scale SM API implemented. A request will be made when `onTick` gets triggered to test the request until the cache flush logic is implemented.


singleton-service/src/service/report.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl<'a> Report<'a> {
+    pub fn service_id(&self) -> &String {
+        &self.service_id
+    }
+
+    pub fn service_token(&self) -> &String {
+        &self.service_token
+    }
```
unleashed, line 27: Your usage (below) of these two methods is only really needing the pointer to the string data (the string slice), not the whole `String` struct by reference. There is a subtle distinction between `&String` and `&str`, the former allows you to check stuff about the buffer holding the string, such as capacity, and if you needed to mutate it via `&mut String` it would allow you to manipulate the string efficiently in-place, ie. enlarge/shrink it and reallocate it, whereas the latter (which is what you generally want) is useful to just passing the string data around and has a limited set of manipulations in-place (no reallocation, no enlarging/shrinking) when using a `&mut str`.

When you just need to read the string, you typically want to pass `&str` - you'll only see `&String` when you need more specific not really stringy things like the underlying buffer capacity that holds the string.

So if you use `self.service_id.as_str()` or even `&*self.service_id` (`String` implements [`Deref<Target=str>`](https://doc.rust-lang.org/std/string/struct.String.html#impl-Deref), so `&*` first dereferences to `str` then takes a reference to it) you should be just fine and allow the usages you make of it to drop the `as_str()` call.

singleton-service/src/service/report.rs
```diff
@@ -0,0 +1,71 @@
...
+    for (user_key, usage) in report.usages().iter() {
+        let application = Application::from(UserKey::from(user_key.as_str()));
+        let usage = Usage::new(usage);
+        app_usage.push((application, usage))
+    }
+    let txns = app_usage
+        .iter()
+        .map(|au| (Transaction::new(&au.0, None, Some(&au.1), None)))
+        .collect::<Vec<_>>();
+    let extensions = extensions::List::new().no_body().push(Extension::Hierarchy);
```
unleashed, line 63: Note that if you don't really need to use any extensions for the report call, as these two (`no_body` and `hierarchy`) don't apply to `report` (also note that they are contradictory: if you need the hierarchy information, well, it comes in the response body of authorize/authrep endpoints :)

singleton-service/src/service/report.rs
```diff
@@ -0,0 +1,71 @@
...
+    })
+}
+
+// build_report_request creates a request which is of type threescalers Report.
+pub fn build_report_request(report: &Report) -> Result<Request, anyhow::Error> {
+    let creds = Credentials::ServiceToken(ServiceToken::from(report.service_token().as_str()));
+    let svc = Service::new(report.service_id().as_str(), creds);
+    let mut app_usage = vec![];
+    for (user_key, usage) in report.usages().iter() {
+        let application = Application::from(UserKey::from(user_key.as_str()));
```
unleashed, line 55: User keys are just one type of application, you will likely also have AppId (with an optional AppKey), so you might need to either store it as an `Application` itself or if you use Strings you should know whether to convert those first to `UserKey`s or `AppId`s.

singleton-service/src/service/report.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+/// Report method will be used by the cache flush implementation (both cache container limit and periodical)
+/// to create a report, which is the proxy level representation. Then it will be used to build the report request
+/// which is of threescalers Report request type.
+pub fn report<'a>() -> Result<Report<'a>, anyhow::Error> {
+    let metrics = [("hits", "1"), ("hits.79419", "1")].to_vec();
+    let mut usages_map: HashMap<String, Vec<(&'a str, &'a str)>> = HashMap::new();
+    usages_map.insert("46de54605a1321aa3838480c5fa91bcc".to_string(), metrics);
```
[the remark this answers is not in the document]
NomadXD: yeah just a test application id to test the `Report` call. The content of the whole `report()` method will change after the cache flush. Now it's just there for the test http call.

singleton-service/src/service/report.rs
unleashed: I'll correct this - I believe we had one useful extension, but we can learn about it later. Meanwhile you can just have this list empty.

singleton-service/src/service/report.rs
NomadXD: okay will implement this behavior when the input that I get from the cache filter changes. This can be probably handled adding this distinction to the application. for this PR i'm just sending a hard coded request to start off the flushing logic.

singleton-service/src/service/report.rs
NomadXD: fixed

singleton-service/src/service/report.rs
NomadXD: okay fixed

rahulanand16nov (approved):
I don't encourage direct copy even though no matter who wrote it. Rest, Everything looks good to me. Approved.

PR #11, opened 2021-06-09
3scale response handling and rate-limit check implemented

rahulanand16nov:
This PR includes handling of response from 3scale authorize endpoint and check for rate-limit. Please suggest any changes that you think should be made. Note: As getters for structs are still not merged into the 'threescalers' upstream, I have used my fork for now but will be changed as soon as that PR is merged. https://github.com/3scale-rs/threescalers/pull/89
cc @unleashed

### DIFF
--- a/cache-filter/src/filter/http.rs
+++ b/cache-filter/src/filter/http.rs
@@ -19,23 +19,31 @@ use threescale::{
     proxy::cache::{get_application_from_cache, set_application_to_cache},
     structs::*,
 };
-use threescalers::response::{Authorization, UsageReports};
+use threescalers::response::{Authorization, OkAuthorization, UsageReports};
 
 const QUEUE_NAME: &str = "message_queue";
 const VM_ID: &str = "my_vm_id";
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, Clone, thiserror::Error)]
 enum RateLimitedError {
     #[error("overflow due to two duration addition")]
     DurOverflowErr,
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, Clone, thiserror::Error)]
 enum CacheHitError {
     #[error("duration since time later than self")]
     TimeConversionErr(#[from] std::time::SystemTimeError),
     #[error("failure of is_rate_limited error")]
     RateLimitErr(#[from] RateLimitedError),
+}
+
+#[derive(Debug, thiserror::Error)]
+enum AuthResponseError {
+    #[error("failure to follow cache hit flow")]
+    CacheHitErr(#[from] CacheHitError),
+    #[error("conversion from i64 time to u64 duration failed")]
+    NegativeTimeErr,
 }
 
 pub struct CacheFilter {
@@ -178,6 +186,64 @@ impl CacheFilter {
         }
         Ok(false)
     }
+
+    fn handle_auth_response(
+        &mut self,
+        response: &OkAuthorization,
+    ) -> Result<(), AuthResponseError> {
+        // Form application struct from the response
+        let key_split = self.cache_key.split('_').collect::<Vec<_>>();
+        let mut state = HashMap::new();
+        let UsageReports::UsageReports(reports) = response.usage_reports();
+
+        for usage in reports {
+            state.insert(
+                usage.metric.clone(),
+                UsageReport {
+                    period_window: PeriodWindow {
+                        start: Duration::from_secs(
+                            usage
+                                .period_start
+                                .0
+                                .try_into()
+                                .or(Err(AuthResponseError::NegativeTimeErr))?,
+                        ),
+                        end: Duration::from_secs(
+                            usage
+                                .period_end
+                                .0
+                                .try_into()
+                                .or(Err(AuthResponseError::NegativeTimeErr))?,
+                        ),
+                        window: period_from_response(&usage.period),
+                    },
+                    left_hits: usage.current_value,
+                    max_value: usage.max_value,
+                },
+            );
+        }
+
+        let mut hierarchy = HashMap::new();
+        if let Some(metrics) = response.hierarchy() {
+            for (parent, childrens) in metrics.iter() {
+                hierarchy.insert(parent.clone(), childrens.clone());
+            }
+        }
+
+        let app = Application {
+            app_id: key_split[0].to_string(),
+            service_id: key_split[1].to_string(),
+            local_state: state,
+            metric_hierarchy: hierarchy,
+        };
+
+        set_application_to_cache(&self.cache_key, &app, true, None);
+
+        match self.handle_cache_hit(&RefCell::new(app)) {
+            Ok(()) => Ok(()),
+            Err(e) => Err(AuthResponseError::CacheHitErr(e)),
+        }
+    }
 }
 
 impl Context for CacheFilter {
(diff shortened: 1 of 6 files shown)


NomadXD (approved):
LGTM. Approved

cache-filter/src/filter/http.rs
```diff
@@ -115,36 +118,33 @@ impl CacheFilter {
...
-                // Period window is expired
-                // Update period window using get_new_window for each metric
-                // But first confirm how to deal with time sync between host and 3scale
-                // Reset left_hits to max value
-            }*/
+            if let Some(usage_report) = app.borrow_mut().local_state.get_mut(metric) {
+                let mut period = &mut usage_report.period_window;
+                if period.window != Period::Eternity && period.end < *current_time {
+                    // Taking care of period window expiration
+                    let num_windows = current_time.checked_sub(period.start).unwrap().as_secs()
```
unleashed, line 127: What you might want to do regarding these functions that return potentially an `Option::None` or `Result::Err` is use a library like `thiserror` to define a custom error type per module that can be built using other types of errors - and can also include error descriptions of your own -, and then change the signature of a function to return ie. `Result<bool, RateLimitedError>` and use the `?` operator to cleanly get rid of the unhappy paths. See https://github.com/3scale/threescale-wasm-auth/blob/d207f09084ebed548de49adf2c937a32d0302bf3/src/configuration/operation/decode.rs for an example (and notice the `?` usage in the code to automatically return the Error type declared on top by building it out of whatever the expression the `?` applies to).

cache-filter/src/filter/http.rs
```diff
@@ -161,8 +161,47 @@ impl Context for CacheFilter {
         match self.get_http_call_response_body(0, body_size) {
             Some(bytes) => {
                 match Authorization::from_str(std::str::from_utf8(&bytes).unwrap()) {
-                    Ok(_response) => {
-                        // Handle cache hit here
+                    Ok(Authorization::Ok(response)) => {
```
unleashed, line 164: Just a heads up: this will change in the 0.8 release of the library, although it will likely have minimal impact.

cache-filter/src/filter/http.rs
```diff
@@ -161,8 +161,47 @@ impl Context for CacheFilter {
...
+                        let mut state = HashMap::new();
+                        let UsageReports::UsageReports(reports) = response.usage_reports();
+                        for usage in reports {
+                            state.insert(
+                                usage.metric.clone(),
+                                UsageReport {
+                                    period_window: PeriodWindow {
+                                        // i64 to u64 but value will always be positive
+                                        start: Duration::from_secs(
+                                            usage.period_start.0.try_into().unwrap(),
```
unleashed, line 176: Same thing as above, you can use `thiserror`, define your module's error type, change the signature to return a `Result` and do `usage.period_start.0.try_into()?,`.

cache-filter/src/filter/http.rs
```diff
@@ -161,8 +161,47 @@ impl Context for CacheFilter {
...
+                                        window: period_from_response(&usage.period),
+                                    },
+                                    left_hits: usage.current_value,
+                                    max_value: usage.max_value,
+                                },
+                            );
+                        }
+                        let mut hierarchy = HashMap::new();
+                        if let Some(metrics) = response.hierarchy() {
+                            for (parent, childrens) in metrics.iter() {
```
unleashed, line 190: typo: `children` (already plural) :)

cache-filter/src/filter/http.rs
```diff
@@ -161,8 +161,47 @@ impl Context for CacheFilter {
...
+                            service_id: key_split[1].to_string(),
+                            local_state: state,
+                            metric_hierarchy: hierarchy,
+                        };
+                        set_application_to_cache(&self.cache_key, &app, true, None);
+                    }
+                    Ok(Authorization::Denied(denied_auth)) => {
+                        info!("Authorization was denied with code: {}", denied_auth.code());
+                        request_process_failure(self, self);
                     }
```
unleashed, line 205: This case currently identifies authorization call failure with denial of authorization, which is a bug. Should be fixed in 0.8 and soon in master, but will require you to adapt this code a bit.

cache-filter/src/utils.rs
```diff
@@ -106,6 +106,18 @@ pub fn do_auth_call<C: proxy_wasm::traits::HttpContext>(
...
+    match res_period {
+        ResponsePeriod::Minute => Period::Minute,
+        ResponsePeriod::Hour => Period::Hour,
+        ResponsePeriod::Day => Period::Day,
+        ResponsePeriod::Week => Period::Week,
+        ResponsePeriod::Month => Period::Month,
+        ResponsePeriod::Year => Period::Year,
+        ResponsePeriod::Eternity => Period::Eternity,
+    }
 }
```
unleashed, line 123: Have you tried to `impl<'a> From<&'a ResponsePeriod> for Period { ... }`? The you can simply have stuff like:

```rust
let response_period: ResponsePeriod = [...];
// get a Period out of the ResponsePeriod (without consuming it, ie. by reference)
let period = Period::from(&response_period);
```

I'd also probably place this conversion next to the place where `Period` is defined.

cache-filter/src/filter/http.rs
rahulanand16nov: Fixed

cache-filter/src/filter/http.rs
rahulanand16nov: Fixed

cache-filter/src/filter/http.rs
rahulanand16nov: As I was writing the patch for `threescale-wasm-auth`, I noticed how simple this pattern is and used it extensively inside the PR. Also, changed the code to use this pattern.

NomadXD (approved):
Can merge if @unleashed is okay with the requested changes.

cache-filter/src/filter/http.rs
```diff
@@ -16,11 +19,33 @@ use threescale::{
...
-use threescalers::response::{Authorization, UsageReports};
+use threescalers::response::{Authorization, OkAuthorization, UsageReports};
 
 const QUEUE_NAME: &str = "message_queue";
 const VM_ID: &str = "my_vm_id";
 
+#[derive(Debug, Clone, thiserror::Error)]
+enum RateLimitedError {
+    #[error("overflow due to two duration addition")]
+    DurOverflowErr,
```
unleashed, line 30: Note that because this is part of an `RateLimitError` enum, it's already clear it is an error, so instead of economizing on the important subject, the Duration, I'd suggest you do so by removing the `Err`/`Error` suffix like this, which leaves no doubt about what has been overflown.
```suggestion
    DurationOverflow,
```

You can also apply this to other error types below.

cache-filter/src/filter/http.rs
```diff
@@ -94,16 +123,13 @@ impl CacheFilter {
...
-            Err(e) => {
-                debug!("Failed to get time from host due to: {}", e);
-                return false;
-            }
-        };
-        if self.is_rate_limited(app, &current_time) {
+
+        let current_time = self.get_current_time().duration_since(UNIX_EPOCH)?;
+
+        let rate_limited: bool = self.is_rate_limited(app, &current_time)?;
```
unleashed, line 131: Indeed much nicer code enabled by the error types and the `?` operator :)

cache-filter/src/filter/http.rs
```diff
@@ -148,74 +184,102 @@ impl CacheFilter {
...
+            Err(e) => Err(AuthResponseError::CacheHitErr(e)),
+        }
     }
 }
 
 impl Context for CacheFilter {
     fn on_http_call_response(&mut self, token_id: u32, _: usize, body_size: usize, _: usize) {
         info!(
-            "ctxt {}: Recieved response from 3scale: token: {}",
+            "ctxt {}: recieved response from 3scale: token: {}",
```
unleashed, line 252: typo `received` :D

unleashed (approved):
Looking good

PR #13, opened 2021-06-12
changed metric value to u64

NomadXD:
changed type of metric to u64


NomadXD:
merging this as a review is not required for the quick fix of the data type.

PR #14, opened 2021-06-12
Fetched inputs from request headers and code restructured

rahulanand16nov:
This PR completes the basic cache filter requirements as I've added the code to fetch inputs from the request headers and restructured for easy type identification from the singleton side of keys. Please let me know if you find any mistakes that I overlooked. Thanks! :)

Note: Only the last commit is relevant because reset of them came automatically as I updated my local fork during the creation of this PR.

### DIFF
--- a/threescale/src/structs.rs
+++ b/threescale/src/structs.rs
@@ -5,12 +5,12 @@ use std::time::Duration;
 
 #[derive(Serialize, Deserialize, Clone, PartialEq)]
 pub enum Period {
-    Minute = 60,
-    Hour = 3600,
-    Day = 86400,
-    Week = 604800,
-    Month = 2592000,
-    Year = 31536000,
+    Minute,
+    Hour,
+    Day,
+    Week,
+    Month,
+    Year,
     Eternity,
 }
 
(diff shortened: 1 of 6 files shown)


threescale/src/structs.rs
```diff
@@ -45,11 +45,168 @@ pub struct UsageReport {
...
+
+impl From<UserKey> for AppIdentifier {
+    fn from(u: UserKey) -> Self {
+        AppIdentifier::UserKey(u)
+    }
+}
+
+impl AppIdentifier {
+    // This cannot return a reference because app_id:app_key needs to be generated
+    pub fn as_string(&self) -> String {
```
NomadXD, line 182: Can you return the type of `AppIdentifier` as well ? Because when consuming these data and creating threescalers application , we need to know which type of key we have.

NomadXD (changes requested):
Requested changes for `AppIdentifier` to return it's type. Also squash and merge to remove unrelated commits.

NomadXD (approved):

rahulanand16nov:
I am creating a new PR without previous commits that are already merged. I'll close this one and apply the patch consisting of the last commit.

PR #15, opened 2021-06-13
Fetched inputs from request headers and code restructured

rahulanand16nov:
This PR completes the basic cache filter requirements as I've added the code to fetch inputs from the request headers and restructured for easy type identification from the singleton side of keys. Please let me know if you find any mistakes that I overlooked. Thanks! :)

### DIFF
--- a/singleton-service/src/service/proxy.rs
+++ b/singleton-service/src/service/proxy.rs
@@ -16,7 +16,7 @@ use thiserror::Error;
 use threescale::upstream::*;
 use threescale::{
     proxy::cache::{get_application_from_cache, set_application_to_cache},
-    structs::{Message, ThreescaleData},
+    structs::{AppIdentifier, CacheKey, Message, ServiceId, ServiceToken, ThreescaleData, UserKey},
     utils::update_metrics,
 };
 use threescalers::http::Request;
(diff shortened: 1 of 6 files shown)


NomadXD (approved):

NomadXD (approved):
Reviewed the PR #14 which is basically the same. Let's wait for @unleashed and then merge after his review.

threescale/src/structs.rs
```diff
@@ -45,11 +45,168 @@ pub struct UsageReport {
...
+}
+
+impl From<UserKey> for AppIdentifier {
+    fn from(u: UserKey) -> Self {
+        AppIdentifier::UserKey(u)
+    }
+}
+
+impl AppIdentifier {
+    // This cannot return a reference because app_id:app_key needs to be generated
```
unleashed, line 181: :+1:

threescale/src/structs.rs
```diff
@@ -45,11 +45,168 @@ pub struct UsageReport {
...
+    }
+
+    pub fn appid_from_str(s: &str) -> AppIdentifier {
+        let v: Vec<&str> = s.split(':').collect();
+        if v.len() == 1 {
+            return AppIdentifier::AppId(AppId(v[0].to_owned()), Some(AppKey(v[1].to_owned())));
+        }
+        AppIdentifier::AppId(AppId(v[0].to_owned()), None)
+    }
+}
```
unleashed, line 203: When you generate a `Self` (ie. AppIdentifier) in an impl block, consider whether you would be best served by implementing `From`, `TryFrom`, or `FromStr`, which have the benefits of working in more scenarios because they are well known traits for which people provide more ergonomic helpers and generic conversions, such as being able to call `"app_id"::parse::<AppIdenfitier>()` if you `impl FromStr`, and being able to specify bounds like `A: Into<AppIdentifier>` if you impl `From<...> for AppIdentifier` (as any `From<Src> for Target` automatically impls `Into<Target> for Src`, though not the other way around).

singleton-service/src/service/proxy.rs
```diff
@@ -153,10 +153,11 @@ impl RootContext for SingletonService {
...
         .collect();
         let threescale1 = ThreescaleData {
-            app_id: "46de54605a1321aa3838480c5fa91bcc".to_string(),
-            service_id: "2555417902188".to_string(),
-            service_token: "6705c7d02e9a899d4db405dc1413361611e4250dfd12ec3dcbcea8c3de7cdd29"
-                .to_string(),
+            app_id: AppIdentifier::UserKey(UserKey::from("46de54605a1321aa3838480c5fa91bcc")),
+            service_id: ServiceId::from("2555417902188"),
+            service_token: ServiceToken::from(
+                "6705c7d02e9a899d4db405dc1413361611e4250dfd12ec3dcbcea8c3de7cdd29",
```
unleashed, line 159: These are still hanging around here, let's try and find a way around hardcoding them.

NomadXD: @unleashed will get fixed from my next PR.

PR #16, opened 2021-06-15
Auth call and app identifier changes for singleton service

NomadXD:
This PR includes,

- cache update operation by performing http calls to 3scale SM API per application. 
- delta store modified to work with app identifier.
- singleton service configuration splitted into several sub configs.
- report and auth call modified to work with app identifier types. (user_key, app_id , app_id+app_key)


NomadXD:
Failing due to threescalers fork. Will send another PR after https://github.com/3scale-rs/threescalers/pull/89 gets merged.

PR #18, opened 2021-06-15
Authorize call and app identifier changes for singleton service

NomadXD:
This PR includes,

- cache update operation by performing http calls to 3scale SM API per application. 
- delta store modified to work with app identifier.
- singleton service configuration splitted into several sub configs.
- report and auth call modified to work with app identifier types. (user_key, app_id , app_id+app_key)
- removed threescalers fork and added threescalers main

### DIFF
--- a/cache-filter/src/filter/http.rs
+++ b/cache-filter/src/filter/http.rs
@@ -16,7 +16,7 @@ use threescale::{
     proxy::cache::{get_application_from_cache, set_application_to_cache},
     structs::*,
 };
-use threescalers::response::{Authorization, OkAuthorization, UsageReports};
+use threescalers::response::{Authorization, AuthorizationStatus, UsageReports};
 
 const QUEUE_NAME: &str = "message_queue";
 const VM_ID: &str = "my_vm_id";
@@ -41,6 +41,8 @@ enum AuthResponseError {
     CacheHitErr(#[from] CacheHitError),
     #[error("conversion from i64 time to u64 duration failed")]
     NegativeTimeErr,
+    #[error("usage reports from 3scale auth response are missing")]
+    UsageNotFound,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -181,6 +183,7 @@ impl CacheFilter {
                     // reset left hits back to max value
                     usage_report.left_hits = usage_report.max_value;
 
+                    // TODO : Use the update_metric()
                     if usage_report.left_hits < *hits {
                         return Ok(true);
                     }
@@ -196,11 +199,13 @@ impl CacheFilter {
 
     fn handle_auth_response(
         &mut self,
-        response: &OkAuthorization,
+        response: &AuthorizationStatus,
     ) -> Result<(), AuthResponseError> {
         // Form application struct from the response
         let mut state = HashMap::new();
-        let UsageReports::UsageReports(reports) = response.usage_reports();
+        let UsageReports::UsageReports(reports) = response
+            .usage_reports()
+            .ok_or(AuthResponseError::UsageNotFound)?;
 
         for usage in reports {
             state.insert(
@@ -292,7 +297,7 @@ impl Context for CacheFilter {
         match self.get_http_call_response_body(0, body_size) {
             Some(bytes) => {
                 match Authorization::from_str(std::str::from_utf8(&bytes).unwrap()) {
-                    Ok(Authorization::Ok(response)) => {
+                    Ok(Authorization::Status(response)) => {
                         if let Err(e) = self.handle_auth_response(&response) {
                             warn!(
                                 "ctxt {}: handling auth response failed: {:?}",
@@ -301,7 +306,7 @@ impl Context for CacheFilter {
                             request_process_failure(self, self)
                         }
                     }
-                    Ok(Authorization::Denied(denied_auth)) => {
+                    Ok(Authorization::Error(denied_auth)) => {
                         info!("authorization was denied with code: {}", denied_auth.code());
                         request_process_failure(self, self);
                         return;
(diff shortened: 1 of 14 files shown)


cache-filter/src/filter/http.rs
```diff
@@ -301,7 +306,7 @@ impl Context for CacheFilter {
                             request_process_failure(self, self)
                         }
                     }
-                    Ok(Authorization::Denied(denied_auth)) => {
+                    Ok(Authorization::Error(denied_auth)) => {
                         info!("authorization was denied with code: {}", denied_auth.code());
                         request_process_failure(self, self);
                         return;
```
unleashed, line 312: This is not a denied authorization, but an error in the authorization call - ie. missing application identifier/user_key, missing service identifier or wrong service token, specified a missing metric, etc.

cache-filter/src/filter/http.rs
```diff
@@ -292,7 +297,7 @@ impl Context for CacheFilter {
         match self.get_http_call_response_body(0, body_size) {
             Some(bytes) => {
                 match Authorization::from_str(std::str::from_utf8(&bytes).unwrap()) {
-                    Ok(Authorization::Ok(response)) => {
+                    Ok(Authorization::Status(response)) => {
                         if let Err(e) = self.handle_auth_response(&response) {
```
unleashed, line 301: Note that you are passing in the `AuthorizationStatus` and `handle_auth_response` just goes on to pick the usage reports, but does not actually check that the authorization is accepted/authorized or rejected/denied.

cache-filter/src/filter/http.rs
NomadXD: ops. last moment change of the threescalers api. will fix that info log.

cache-filter/src/filter/http.rs
NomadXD: @unleashed This is actually on the cache filter side. I didn't change anything related to the implementation. Just changed the Authorization to use the new API. eg: `Authorization::Ok`  to `Authorization::Status`. And yes I think `handle_auth_response` should take account status as well. We will get both 200 and 409 through that `Authorization::Ok`. `handle_auth_response` in singleton service takes status into account. https://github.com/NomadXD/gsoc-wasm-filters/blob/dev/auth-call/singleton-service/src/service/proxy.rs#L252
Will fix this in a separate PR. cc @rahulanand16nov

rahulanand16nov: @NomadXD Better do it in this PR as it's a small change - let me know if you need any help with it.

cache-filter/src/filter/http.rs
rahulanand16nov: Also, change the variable name above it.

singleton-service/src/service/auth.rs
```diff
@@ -0,0 +1,86 @@
...
+    service::Service,
+    transaction::Transaction,
+};
+
+/// Proxy level representation for authorization request.
+pub struct Auth {
+    pub service_id: String,
+    pub service_token: String,
+    pub app_id: AppIdentifier,
+}
```
rahulanand16nov, line 18: You are publishing and providing getters as well. Have either not both.

singleton-service/src/service/auth.rs
```diff
@@ -0,0 +1,86 @@
...
+            debug!("AppIdentifier AppId : {}", auth.app_id().as_string());
+            app = Application::from_app_id(auth.app_id().as_string())
+        }
+        AppIdentifier::AppId(_app_id, _app_key) => {
+            let app_key = auth.app_id().as_string();
+            debug!("AppIdentifier AppId+AppKey : {}", app_key);
+            let keys = app_key.split('_').collect::<Vec<_>>();
+            app = Application::from_app_id_and_key(keys[0], keys[1])
+        }
+    }
```
rahulanand16nov, line 75: Can you please shed some light on the following concerns:
- You've pattern-matched `user_key` but not used it instead provided app_id in place of it?
- Even in the second pattern? you have required fields but still reforming them using `as_string()`?
- Same goes for 3rd pattern.
Maybe I am getting this wrong?

singleton-service/src/service/report.rs
```diff
@@ -64,10 +64,26 @@ pub fn build_report_request(report: &Report) -> Result<Request, anyhow::Error> {
...
+                debug!("AppIdentifier AppId : {}", app_id.as_string());
+                app = Application::from_app_id(app_id.as_string())
+            }
+            AppIdentifier::AppId(_app_id, _app_key) => {
+                let app_key = app_id.as_string();
+                debug!("AppIdentifier AppId+AppKey : {}", app_key);
+                let keys = app_key.split('_').collect::<Vec<_>>();
+                app = Application::from_app_id_and_key(keys[0], keys[1])
+            }
+        }
```
rahulanand16nov, line 84: Duplicate code with the same concerns mentioned above.

rahulanand16nov (changes requested):

cache-filter/src/filter/http.rs
NomadXD: fixed

cache-filter/src/filter/http.rs
NomadXD: fixed

singleton-service/src/service/auth.rs
NomadXD: fixed

singleton-service/src/service/auth.rs
NomadXD: sorry this was a temp thing i was trying and forgot to fix it before the PR. fixed

singleton-service/src/service/report.rs
NomadXD: fixed

rahulanand16nov (approved):
LGTM now :)

PR #19, opened 2021-06-21
Cache update, singleton cache integration and few other fixes

NomadXD:
This PR includes,

- Cache update from the singleton using authorize response.
- Minor fixes for cache filter
- Cache filter , singleton integration (Currently working and tested for `app_id` pattern. `user_key` is not supported until we have the mapping from `user_key: app_id`)
- Deployments folder to run a cache and singleton with docker-compose.
- Custom hash functions implemented for `CacheKey` and `AppIdentifier`
- Cache filter Auth call for different AppIdentifier types


Makefile
```diff
@@ -0,0 +1,28 @@
...
+		cargo build --target=wasm32-unknown-unknown --release
+		cp target/wasm32-unknown-unknown/release/cache_filter.wasm ./deployments/docker-compose/cache_filter.wasm
+		cp target/wasm32-unknown-unknown/release/singleton_service.wasm ./deployments/docker-compose/singleton_service.wasm
+
+.PHONY: clean
+clean:
+		rm ./deployments/docker-compose/cache_filter.wasm
+		rm ./deployments/docker-compose/singleton_service.wasm
+
+
```
NomadXD, line 28: Will fix

threescale/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(clippy::all, clippy::cargo)]
+#![allow(clippy::multiple_crate_versions)]
```
NomadXD, line 2: @unleashed Had to add this because there was conflict between dependencies of `threescalers` and `rust-url` and clippy doesn't pass. 
https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions

unleashed: You should not need this if you patch the dependency in the main Cargo.toml for the workspace with something like:

```toml
[patch.crates-io]
percent-encoding = { git = "https://github.com/3scale-rs/rust-url", branch = "3scale" }
```

threescale/src/structs.rs
```diff
@@ -181,14 +181,7 @@ impl AppIdentifier {
...
-            AppIdentifier::AppId(AppId(id), key) => {
-                let mut res: String = id.clone();
-                if let Some(AppKey(app_key)) = key {
-                    res.push(':');
-                    res.push_str(app_key.as_str());
-                }
-                res
-            }
+            AppIdentifier::AppId(AppId(id), _key) => id.clone(),
             AppIdentifier::UserKey(UserKey(user_key)) => user_key.clone(),
```
unleashed: Since you are no longer generating the format "app_id:app_key" then the comment no longer applies - so either it is now wrong and must be deleted or we can now switch this to use a string slice `&str`.

cache-filter/src/filter/http.rs
```diff
@@ -96,7 +100,7 @@ impl HttpContext for CacheFilter {
                 info!("ctxt {}: Cache Miss", context_id);
                 // TODO: Avoid multiple calls for same application
                 // saving request data to use when there is response from 3scale
-                self.req_data = request_data.clone();
+                // self.req_data = request_data.clone();
```
rahulanand16nov, line 103: You need this data to be present when the response comes from 3scale.

threescale/src/utils.rs
```diff
@@ -1,6 +1,20 @@
...
+    match res_period {
+        ResponsePeriod::Minute => Period::Minute,
+        ResponsePeriod::Hour => Period::Hour,
+        ResponsePeriod::Day => Period::Day,
+        ResponsePeriod::Week => Period::Week,
+        ResponsePeriod::Month => Period::Month,
+        ResponsePeriod::Year => Period::Year,
+        ResponsePeriod::Eternity => Period::Eternity,
+    }
+}
```
rahulanand16nov, line 20: Can you try to implement `from` trait as it was mentioned by Alex in my last (or 2nd last PR?). I had issues with dependency conflicts back then used this instead. Now, that it can be placed into threescale, please convert this into a `from` trait for `Period` Enum. Thanks!

cache-filter/src/filter/http.rs
```diff
@@ -79,13 +82,14 @@ impl HttpContext for CacheFilter {
...
         self.cache_key = CacheKey::from(&request_data.service_id, &request_data.app_id);
+        self.req_data = request_data.clone();
         match get_application_from_cache(&self.cache_key.as_string()) {
             Some((app, _)) => {
                 info!("ctxt {}: Cache Hit", context_id);
 
                 let app_ref = RefCell::new(app);
                 match self.handle_cache_hit(&app_ref) {
-                    Ok(()) => Action::Pause,
+                    Ok(()) => Action::Continue,
```
rahulanand16nov, line 92: Nice find.

singleton-service/src/service/auth.rs
```diff
@@ -72,8 +77,9 @@ pub fn build_auth_request(auth: &Auth) -> Result<Request, anyhow::Error> {
         }
     }
     let txn = vec![(Transaction::new(&app, None, None, None))];
-    // TODO : Enable list keys extension.
-    let extensions = extensions::List::new().push(extensions::Extension::Hierarchy);
+    let extensions = extensions::List::new()
+        .push(extensions::Extension::Hierarchy)
+        .push(extensions::Extension::ListAppKeys(1.to_string().into()));
```
unleashed, line 82: You should be able to use `"1".into()`, since `&'static str` can become a `Cow<'static, str>` (and reference static memory rather than allocate a new string).

cache-filter/src/filter/http.rs
```diff
@@ -206,7 +210,11 @@ impl CacheFilter {
         let UsageReports::UsageReports(reports) = response
             .usage_reports()
             .ok_or(AuthResponseError::UsageNotFound)?;
-
+        let app_keys = response
+            .app_keys()
+            .ok_or(AuthResponseError::ListKeysMissing)?;
+        let app_id = AppIdentifier::from(AppId::from(app_keys.app_id().unwrap().as_ref()));
+        let service_id = ServiceId::from(app_keys.service_id().unwrap().as_ref());
```
rahulanand16nov, line 217: In future, those `unwraps` can become unsafe, how about an error enum for them as well?

cache-filter/src/filter/http.rs
```diff
@@ -241,17 +249,37 @@ impl CacheFilter {
...
-            local_state: state,
-            metric_hierarchy: hierarchy,
-        };
+        let app;
+        if let Some(app_keys) = response.app_keys() {
+            let keys = app_keys
+                .keys()
+                .iter()
+                .map(|app_key| AppKey::from(app_key.as_ref()))
+                .collect::<Vec<_>>();
```
rahulanand16nov, line 258: You have made sure that `response.app_key()` contains something at line number: 215. You can straight-away access the inner structure.

cache-filter/src/filter/http.rs
```diff
@@ -241,17 +249,37 @@ impl CacheFilter {
...
+                app_keys: None,
+            };
+        }
 
         set_application_to_cache(&self.cache_key.as_string(), &app, true, None);
 
         match self.handle_cache_hit(&RefCell::new(app)) {
-            Ok(()) => Ok(()),
+            Ok(()) => {
+                self.resume_http_request();
```
rahulanand16nov, line 280: Another great find! btw, this should be a part of `handle_cache_hit` flow. Can you please shift this to after line: 154 i.e if MQ doesn't fail.

cache-filter/src/utils.rs
```diff
@@ -33,31 +39,32 @@ pub fn do_auth_call<C: proxy_wasm::traits::HttpContext>(
...
-        request_data.app_id.as_string().as_ref(),
-    );
+    let app;
+    match &request_data.app_id {
+        AppIdentifier::UserKey(user_key) => app = Application::from_user_key(user_key.as_ref()),
+        AppIdentifier::AppId(app_id, None) => app = Application::from_app_id(app_id.as_ref()),
+        AppIdentifier::AppId(app_id, Some(app_key)) => {
+            app = Application::from_app_id_and_key(app_id.as_ref(), app_key.as_ref())
+        }
+    }
```
rahulanand16nov, line 49: Looks great! Just remove the comment as it's not needed anymore. :)

rahulanand16nov (changes requested):
Looks great! Just minor changes request.

cache-filter/src/filter/http.rs
rahulanand16nov: Please refer to the convo in the slack.

threescale/src/lib.rs
NomadXD: Okay great !! It worked.

threescale/src/structs.rs
NomadXD: fixed

singleton-service/src/service/auth.rs
NomadXD: fixed for all the places

cache-filter/src/filter/http.rs
NomadXD: it's actually not removed. when you place it inside a match arm , request data binding happens only for that. so in the case of cache hit , request data is not getting binded to the HttpContext and I'm getting empty data. So I just moved the request data binding to the outer scope.

threescale/src/utils.rs
NomadXD: fixed

cache-filter/src/filter/http.rs
NomadXD: fixed

cache-filter/src/filter/http.rs
NomadXD: yeah my bad. didn't notice it.

cache-filter/src/filter/http.rs
NomadXD: yeah i was actually going to put that there but in the case of a cache hit , it returns `Action::Continue` from the outer function. So that's why I handled it from outside (not to use `self.resume` and `Action::Continue` at the same time). But anyway there's no issue with using both `resume` and `continue` , it will take whoever comes first. For now I moved it into `handle_cache_hit` and maybe we can refactor things later.

cache-filter/src/utils.rs
NomadXD: fixed

rahulanand16nov (approved):

PR #20, opened 2021-06-23
Requiring conversion: UserKey to AppId for every request

rahulanand16nov:
This PR includes:
- Storing apps in the cache with `serviceid_appid` format only.
- Simplification of `proxy` module inside `threescale` module.


cache-filter/src/filter/http.rs
```diff
@@ -83,7 +85,26 @@ impl HttpContext for CacheFilter {
...
+                    self.req_data.app_id = request_data.app_id.clone();
+                    self.cache_key.set_app_id(&request_data.app_id);
+                }
+                Err(e) => {
+                    info!(
+                        "ctxt {}: failed to map user_key to app_id: {:?}",
+                        context_id, e
+                    );
+                    // TODO: avoid multiple calls for identical requests
+                    return do_auth_call(self, self, &request_data);
```
unleashed, line 102: This case will happen when we don't know about a specific user key - it might or might not exist, so perhaps it is valid once we fetch the data from an auth call. I'd lower the logging to debug level.

deployments/docker-compose/Dockerfile
```diff
@@ -4,4 +4,4 @@ COPY ./envoy.yaml /etc/envoy/envoy.yaml
 COPY ./cache_filter.wasm /usr/local/bin/cache_filter.wasm
 COPY ./singleton_service.wasm /usr/local/bin/singleton_service.wasm
 RUN chmod go+r /etc/envoy/envoy.yaml /usr/local/bin/cache_filter.wasm /usr/local/bin/singleton_service.wasm
-CMD /usr/local/bin/envoy -c /etc/envoy/envoy.yaml
+CMD /usr/local/bin/envoy -c /etc/envoy/envoy.yaml -l debug
```
NomadXD, line 7: Let's keep the default log levels as it and change them whenever we need them accordingly in development. (debug, trace and component level logging).

NomadXD (changes requested):

deployments/docker-compose/Dockerfile
rahulanand16nov: It was not supposed to be pushed. fixed :)

PR #21, opened 2021-06-25
Increasing guarantees and updating applications to account for consumption

rahulanand16nov:
This PR includes:
- Increasing method parameter type guarantees and avoiding useless error handling.
- `limit_check_and_update_application` implemented.
- Fixed edge case when the request is not authorized but data like usage_reports and hierarchy is available in the auth response.
- Moved from trait for `Period` to its expected place.

### DIFF
--- a/threescale/src/utils.rs
+++ b/threescale/src/utils.rs
@@ -1,4 +1,4 @@
-use crate::structs::{Application, Period, ThreescaleData};
+use crate::structs::{Application, Hierarchy, Metrics, Period, ThreescaleData};
 use std::time::Duration;
 
 #[derive(Debug, Clone, thiserror::Error)]
@@ -51,4 +51,17 @@ pub fn limit_check_and_update_application(
         }
     }
     Ok(())
+}
+
+// It takes the provided hierarchy structure, and uses it
+// to determine how the metrics, m, are affected, incrementing parent metrics
+// based on the value of the parents child/children metrics.
+pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
+    for (parent, children) in hierarchy.iter() {
+        for (metric, hits) in metrics.borrow().iter() {
+            if children.contains(metric) {
+                *metrics.borrow_mut().entry(parent.to_string()).or_insert(0) += *hits;
+            }
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


threescale/src/utils.rs
```diff
@@ -1,23 +1,54 @@
...
+                    .ok_or(UpdateMetricsError::DurationOverflow)?;
+
+                // reset left hits back to max value
+                usage_report.left_hits = usage_report.max_value;
+            }
+
+            if usage_report.left_hits < *hits {
+                return Err(UpdateMetricsError::RateLimited);
+            }
+            usage_report.left_hits -= *hits;
```
NomadXD, line 50: I'm not very sure about updating the usages this way. I thought we'll need a more complex logic with the hierarchies as well. I went through the section that the auth-filter([https://github.com/3scale/threescale-wasm-auth/blob/4dfcf9ae9e9b6d9651880c3c38eef8082ef6f0f5/src/proxy/authrep.rs#L90()) sets usages and seems like we can straightaway update the usages like that since auth-filter maps the path with usage rules and sends us only the metrics related to this particular url. But I'm not sure whether we need to handle scenarios like `update parent only` or `update child only`. I remember something like that from previous conversations.

rahulanand16nov: My bad! I forgot to implement `add_hierarchy_to_metrics` before pushing that code; It's now up and you can review it.

threescale/src/utils.rs
```diff
@@ -1,23 +1,70 @@
...
+// to determine how the metrics, m, are affected, incrementing parent metrics
+// based on the value of the parents child/children metrics.
+pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
+    for (parent, children) in hierarchy.iter() {
+        for (metric, hits) in metrics.borrow().iter() {
+            if children.contains(metric) {
+                *metrics
+                    .borrow_mut()
+                    .entry(parent.to_string())
+                    .or_insert(*hits) += *hits;
```
unleashed, line 66: Hm, if the entry doesn't exist it will already have `hits`, so you'd be adding 2x hits. Probably want `.or_insert(0)` instead.

rahulanand16nov: Fixed

NomadXD (approved):
LGTM

PR #24, opened 2021-06-28
integration tests, docs, auth filter added

NomadXD:
This PR includes,

- Integration of `threescale-wasm-auth` filter as a git sub-module
- Integration test framework for writing integration tests
- Integration tests for several scenarios
- Makefile build, test, run commands
- Updated README with build, test, run commands and few examples
- Added integration tests to github actions as a dependent job of the build job


.github/workflows/ci.yaml
```diff
@@ -45,6 +45,76 @@ jobs:
...
+          go version
+          go get -u golang.org/x/lint/golint
+          go get github.com/stretchr/testify/assert
+
+      - name: Building application
+        run: make build
+
+      - name: Running Integration tests
+        run: make integration    
+
```
NomadXD, line 120: will fix

Makefile
```diff
@@ -1,27 +1,73 @@
...
+	cp deployments/docker-compose/singleton_service.wasm integration-tests/artifacts/singleton_service.wasm
+	cp deployments/docker-compose/threescale_wasm_auth.wasm integration-tests/artifacts/threescale_wasm_auth.wasm
+	go clean -testcache
+	go test -p 1 ./... -v
+	rm -rf integration-tests/artifacts
 
+run:
+	@echo "> Starting services"
+	docker-compose -f deployments/docker-compose/docker-compose.yaml up --build
+
```
NomadXD, line 73: will fix

.github/workflows/ci.yaml
```diff
@@ -45,6 +45,76 @@ jobs:
...
+        with:
+          crate: wasm-snip
+          version: latest
+      - name: Installing binaryen
+        run: sudo apt-get install binaryen
+        
+      - name: Setting up golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.0'
```
unleashed, line 107: Probably not very important, but note you are using 1.16 elsewhere in the go module configuration.

integration-tests/app_id_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	res, errHTTP := client.Do(req)
+	if errHTTP != nil {
+		fmt.Printf("Error: %v", errHTTP)
+		BuildStopContainers()
+		t.Fatalf("Error sending http request: %v", errHTTP)
+	}
+	fmt.Printf("Response: %v", res)
+	assert.Equal(t, 403, res.StatusCode, "Invalid http response code for appId forbidden test: %d", res.StatusCode)
+	assert.Equal(t, "3scale", res.Header.Get("Powered-By"), "Inavlid http header for powered-by: %v", res.Header.Get("Powered-By"))
```
unleashed, line 78: Don't match on this header as it will most likely be removed.

unleashed (commented):
Good job, Lahiru!

Regarding the way we test, while it should be functional, it takes quite a bit of time to run each test. Perhaps we can run the compose setup and tear down and then up just the Envoy container, and in the middle of both steps we can copy or select a different configuration for Envoy. Maybe we can just change the configuration live (but I think this is problematic, not tested as of late though), or we could just a few instances of envoy in the compose setup, each one with a different configuration (provided we only have a handful of configurations tops).

NomadXD:
@unleashed I was thinking to have something like a test suite and group integration tests to run on different configurations. So that we don't have to start and stop docker containers for each test but instead start and stop for a group of related tests. Golang native test package doesn't have that but can be implemented with few helpers or using testify. And we can try out your suggestions also. As you said better to not use dynamic config from a file since it doesn't always work. And can use different listeners for tests that doesn't depend on the singleton state. (because singleton instance will be the same even we use different listeners). On a side note, testing the behavior of the cache filter is fairly straightforward when compared with the singleton because it doesn't provide a noticeable response/output to client. One thing we can do is add logs and output them to `stdout` and filter that for the expected for some action.

.github/workflows/ci.yaml
NomadXD: Sorry I have several golang versions installed on my machine and was working with 1.16. Will change the CI to 1.16.

integration-tests/app_id_test.go
NomadXD: Okay will remove that. Anything else that needs to be checked other than http status ?

unleashed: HTTP rate limiting headers once we have that implemented :)

rahulanand16nov (approved):
LGTM! Great PR :)

PR #25, opened 2021-07-01
re-modification of apps for retries

rahulanand16nov:
This PR adds re-modification of newly retrieved apps when the application set is failed due to contention with other threads.

### DIFF
--- a/cache-filter/src/filter/http.rs
+++ b/cache-filter/src/filter/http.rs
@@ -15,8 +15,8 @@ use std::time::{Duration, UNIX_EPOCH};
 use std::vec;
 use threescale::{
     proxy::{
-        get_app_id_from_cache, get_application_from_cache, set_app_id_to_cache,
-        set_application_to_cache, CacheError, CacheKey,
+        get_app_id_from_cache, get_application_from_cache, set_app_id_to_cache, CacheError,
+        CacheKey,
     },
     structs::*,
     utils::*,
@@ -26,7 +26,7 @@ use threescalers::response::{Authorization, AuthorizationStatus};
 const QUEUE_NAME: &str = "message_queue";
 const VM_ID: &str = "my_vm_id";
 
-#[derive(Debug, Clone, thiserror::Error)]
+#[derive(Debug, thiserror::Error)]
 enum CacheHitError {
     #[error("duration since time later than self")]
     TimeConversionErr(#[from] std::time::SystemTimeError),
@@ -34,6 +34,8 @@ enum CacheHitError {
     MetricsCheckFail(#[from] UpdateMetricsError),
     #[error("unable to resolve message queue with specified vm_id and q_name")]
     MQNotFound,
+    #[error("failed to fetch the app during app_set retries")]
+    AppFetchFail(#[from] CacheError),
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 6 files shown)


NomadXD (commented):
LGTM. Added one comment.

cache-filter/src/filter/http.rs
```diff
@@ -148,37 +150,72 @@ impl CacheFilter {
...
-            Err(UpdateMetricsError::RateLimited) => {
-                info!("ctxt {}: request is rate-limited", self.context_id);
-                // TODO: add how many similar requests can be accepted.
-                self.send_http_response(429, vec![], Some(b"Request rate-limited.\n"));
+                Err(UpdateMetricsError::CacheUpdateFail) => {
+                    info!(
+                        "ctxt {}: try #{} failed to set application to cache",
+                        self.context_id, num_try
+                    );
+                    if num_try < max_tries {
```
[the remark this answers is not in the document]
rahulanand16nov: Actually, change is required in the start `for num_try in 1..(max_tries+1)`. We need to check that because we don't want to get a new application if that's the last try for the set.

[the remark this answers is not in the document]
NomadXD: ah yes. for loop should go up to `max_retries+1` and if is required.

NomadXD:
@rahulanand16nov Integration tests failed because of an issue with the way that we handle the auth response in the cache filter. I changed the product to unlimited plan to prevent the issue with limits. So now as the auth response we get, 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<status>
	<authorized>true</authorized>
	<plan>Unlimited</plan>
	<app_keys app="fcf4db29" svc="2555417902188">
		<key id="9a0435ee68f5d647f03a80480a97a326"/>
	</app_keys>
	<hierarchy>
	</hierarchy>
</status>
```
But in the cache filter here https://github.com/3scale-labs/gsoc-wasm-filters/blob/a050992f6a92cc150994baad5caae63bda3aa796/cache-filter/src/filter/http.rs#L318 , if there is no usage report it is considered as unauthorized and try to send `403`. And there when you try to send `response.reason().unwrap().as_bytes()`, it causes panic because reason is `None`. I think the check should be `authorized OR some usage report` to handle all cases.

NomadXD (approved):
LGTM

cache-filter/src/filter/http.rs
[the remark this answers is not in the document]
unleashed: What would happen if a user sets `max_tries` to the value `4294967296`?

cache-filter/src/filter/http.rs
```diff
@@ -261,9 +298,9 @@ impl CacheFilter {
             app_keys: Some(keys),
         };
 
-        set_application_to_cache(&self.cache_key.as_string(), &app, true, None);
-
-        match self.handle_cache_hit(&mut app) {
+        // note: we have made an assumption that there is not contention with other threads
+        // since it's a fresh application and should not be present in the cache.
```
unleashed, line 302: Would this hold if we launch two parallel requests using the same application?

rahulanand16nov: We are trying to prevent parallel calls so that only a single call to the 3scale is made for a unique application.

unleashed: What is the mechanism used to prevent those potential parallel calls? If that still allowed two potential writers, I'd be extra cautious that they will actually be writing the same exact thing and that no read happens at the same time.

rahulanand16nov: I am working on the POC to avoid those parallel calls but a possible solution is to store context_ids into shared data and avoid a call if there is already a pending request from 3scale for that application.

unleashed: Yes, that might be enough.

PR #27, opened 2021-07-02
default, periodical and container limit flush added and test suite added

NomadXD:
This PR includes,
- default, periodical and container limit flush
- integration test suite


integration-tests/app_id_test.go
```diff
@@ -7,110 +7,116 @@ import (
...
-		StopContainers("./configs/app-id/docker-compose.yaml")
-		t.Fatalf("Error creating http request: %v", errReq)
-	}
+	require.Nilf(suite.T(), errReq, "Error creating the HTTP request: %v", errReq)
 	req.Header = http.Header{
 		"Host":      []string{"localhost"},
-		"x-app-id":  []string{"fcf4db29"},
-		"x-app-key": []string{"9a0435ee68f5d647f03a80480a97a326"},
+		"x-app-id":  []string{"23f118be"},
+		"x-app-key": []string{"44d128988763aee1b0ff0691f9686f7e"},
```
unleashed, line 47: Will definitely need to read these from the environment or some file accessible from these tests.

integration-tests/app_id_test.go
```diff
@@ -7,110 +7,116 @@ import (
...
+	suite.Run(t, new(AppCredentialTestSuite))
+}
+
+func (suite *AppCredentialTestSuite) TearDownSuite() {
+	fmt.Println("Stopping AppCredentialTestSuite")
+	errStop := StopContainers("./configs/app-id/docker-compose.yaml")
+	if errStop != nil {
+		fmt.Printf("Error stopping : %v", errStop)
+	}
+}
```
unleashed, line 38: Not important, but just a minor note for your own thoughts: when you have a method that does not require the self object (ie. `suite` in this case), that usually indicates that either the method should be a free standing function because it has no relation to the object or that the object is lacking some relationship to the method. In some cases there might arguably exist a point about namespacing/grouping together related functionality, but it's usually the case that there's a hidden object to be discovered there.

integration-tests/app_id_test.go
```diff
@@ -7,110 +7,116 @@ import (
...
+	fmt.Printf("Response: %v", res)
+	assert.Equal(suite.T(), 403, res.StatusCode, "Invalid http response code for user_key forbidden test: %v", res.StatusCode)
 }
+
+// func (suite *AppCredentialTestSuite) TestUnlimitedUserKey() {
+// 	client := &http.Client{}
+// 	req, errReq := http.NewRequest("GET", "http://127.0.0.1:9095/", nil)
+// 	require.Nilf(suite.T(), errReq, "Error creating the HTTP request: %v", errReq)
+// 	q := req.URL.Query()
+// 	q.Add("api_key", "10e81d5c065a537b05ab7d78a7156fc5")
```
unleashed, line 101: Is this part of the WIP?

unleashed (commented):
@NomadXD it would be easier for me to go through the changes if you splitted them so the test changes go in a different commit than the filter changes. Check out comments (minor stuff only).

integration-tests/app_id_test.go
NomadXD: @unleashed Encrypted secrets option is there but didn't use it because keys may get changed during the dev stage. Once we are stable with the tests, we can add to that. But we'll need to add several app_ids, service_ids. Is there a better option than encrypted secrets ?

integration-tests/app_id_test.go
NomadXD: yeah. `StartContainers` and `StopContainers` are helper methods that are shared by all the test suites. So didn't implement them as receiver functions.

integration-tests/app_id_test.go
NomadXD: To support unlimited applications, few changes are required in the cache filter side. So just commented out until it gets fixed. Can remove if this looks ugly

unleashed: No, keep it, but either make sure it compiles or if you comment it out then add a comment on top to know why.

integration-tests/app_id_test.go
NomadXD: @unleashed Sorry I was saying something else in the comment. So `TearDownSuite` requires a pointer receiver because it's a method that is defined in the TestSuite interface. So it's something like you mentioned regarding name-spacing and grouping.

rahulanand16nov (approved):

PR #30, opened 2021-07-07
Integration tests + singleton improvements

NomadXD:
This PR includes,
- Handling HTTP timeouts for the HTTP requests from singleton service
- Apisonator, redis, mock backend and envoy under one docker network to run integration tests
- Dynamically adding the created envoy to the docker network of the local services
- Golint issues fixed + added to CI


rahulanand16nov (approved):
LGTM

singleton-service/src/service/proxy.rs
```diff
@@ -210,17 +213,27 @@ impl Context for SingletonService {
             .find(|(key, _)| key.as_str() == ":status")
             .map(|(_, value)| value)
             .unwrap();
-        match self.get_http_call_response_body(0, body_size) {
-            Some(bytes) => {
-                info!("Auth response");
-                // TODO : Handle auth processing.
-                self.handle_auth_response(bytes, &status).unwrap();
+        if status != TIMEOUT_STATUS {
```
unleashed, line 216: Is there no other relevant `50X` status that can happen here (and I'm specifically thinking not just about integration tests but also the real thing)? IOW, what would happen if the response status was `500 Internal Server Error` or 501/502 for some temporary gateway glitch?

unleashed (commented):
LGTM, but check the comment I made about the status check.

singleton-service/src/service/proxy.rs
NomadXD: @unleashed All those `50X` status from upstream service are handled inside the `handle_auth_response` method after deserializing the response bytes. (using threescalers `Authorization::Status` and `Authorization::Error`). And here I'm handling the gateway timeout which is raised from the envoy async http client in case a response from the upstream doesn't arrive before the configured timeout.

unleashed: Ok, sounds good :+1:

PR #32, opened 2021-07-12
Delta store flush based on memory limit and few other fixes

NomadXD:
This PR includes,
- Delta store flush based on memory limit
- Singleton usage report left hits issue fixed
- Cache filter for loop issue fixed

**Brief summary about the memory limit calculation** 

We want to calculate the memory allocation for the following hashmap. 

`HashMap<String, HashMap<AppIdentifier, HashMap<String, u64>>>`. This hashmap holds the services, application and metrics according the following structure. 

```sh
- Service
    - Application
        - Metric : Value
```
There can be several scenarios for the delta store update. 

1. When a new service gets added. It will add entries to all the hashmaps.
In this case for the memory increase, following things are considered.

- Dynamic memory allocation of the entries on the heap.
- Hashmap allocation of keys and values for all the hashmaps (service, apps, metrics) (direct alloc + transitive alloc in the PR)

2.  When a new application gets added. It will add entries to the apps hashmap and metrics hashmap.
In this case for the memory increase, following things are considered.

- Dynamic memory allocation of the entries on the heap.
- Hashmap allocation of keys and values for application hashmap and metrics hashmap (direct alloc + transitive alloc in the PR)

3. When a new metric gets added to a already existing application. 
In this case for the memory increase, following things are considered.

- Dynamic memory allocation of the entries on the heap.
- Hashmap allocation of keys and values for metrics hashmap (direct alloc in the PR)

So in all of the cases dynamic memory allocation for the values on the heap + allocation of the hashmap keys and values are considered. (the value we get is an underestimate due to reasons like control bytes but a good approximation for our use case)

### DIFF
--- a/singleton-service/src/service/deltas.rs
+++ b/singleton-service/src/service/deltas.rs
@@ -3,10 +3,7 @@ use chrono::offset::Utc;
 use chrono::DateTime;
 use log::info;
 use std::collections::HashMap;
-use threescale::{
-    structs::{AppIdentifier, ThreescaleData},
-    utils::*,
-};
+use threescale::structs::{AppIdentifier, ThreescaleData};
 
 /// DeltaStore is an in-memory storage built using nested hashmaps to store deltas for different
 /// metrics related to different applications until they gets flushed. Data stored in delta store
@@ -23,9 +20,15 @@ pub struct DeltaStore {
     //                  - Metric : Value
     pub deltas: HashMap<String, HashMap<AppIdentifier, HashMap<String, u64>>>,
 
-    // Represents the request count. Gets incremented by 1 for each request passing
-    // through the proxy. Used together with the capacity attribute to implement a
-    // container filling mechansim for cache flush in case of high network traffic scenarios.
+    // Represents a value which is proportional to the memory allocation (underestimate appoximation).
+    // Only the memory allocation for deltas hashmap keys and values are considered. Used together with
+    // the capacity attribute to implement a container filling mechansim for cache flush in case of high
+    // network traffic scenarios.
+    // NOTE : Do not interpret this values as the memory allocation of the delta store because only the
+    // allocation of hashmap keys and values are considerd. Dynamic memory allocation and other control
+    // bytes are not considerd here. So not suitable to take decisions for memory management.
+    // Only intended to provide a mechanism for the user to configure a value for delta store flush to
+    // flush cache based on a value propotional to memory allocation.
     pub memory_allocated: usize,
 
     // DeltaStoreConfig contains all the configurations related for delta store.
@@ -72,8 +75,6 @@ impl DeltaStore {
             },
             None => {
                 info!("No service and application found for the given key combination");
-                // Current allocation of the services hashmap before new entry.
-                let prev_alloc = self.deltas.capacity();
                 let mut usages: HashMap<AppIdentifier, HashMap<String, u64>> = HashMap::new();
                 usages.insert(
                     threescale.app_id.clone(),
@@ -84,26 +85,20 @@ impl DeltaStore {
                     threescale.service_id.as_ref(),
                     threescale.service_token.as_ref()
                 );
-                // Entry bytes denotes the dynamic heap allocation for the entry.
-                let entry_bytes = delta_key.capacity()
-                    + threescale.app_id.dynamic_allocated_size()
-                    + threescale.metrics.borrow().dynamic_allocated_size();
+                self.deltas.insert(delta_key, usages);
                 // transitive_alloc denotes the allocations of the inner hashmaps due to
                 // new entry.
-                let transitive_alloc = usages.capacity()
-                    * (std::mem::size_of::<HashMap<String, u64>>()
-                        + std::mem::size_of::<AppIdentifier>())
-                    + threescale.metrics.borrow().capacity()
+                let transitive_alloc = (std::mem::size_of::<HashMap<String, u64>>()
+                    + std::mem::size_of::<AppIdentifier>())
+                    + threescale.metrics.borrow().len()
                         * (std::mem::size_of::<String>() + std::mem::size_of::<u64>());
-                self.deltas.insert(delta_key, usages);
+
                 // new_alloc denotes the new allocation of the services hashmap.
-                let new_alloc = self.deltas.capacity();
-                let direct_alloc = (new_alloc - prev_alloc)
-                    * (std::mem::size_of::<String>()
-                        + std::mem::size_of::<HashMap<AppIdentifier, HashMap<String, u64>>>());
+                let direct_alloc = std::mem::size_of::<String>()
+                    + std::mem::size_of::<HashMap<AppIdentifier, HashMap<String, u64>>>();
                 // Total value of the delta store memory allocation increase is equal to direct_alloc of the
-                // services hashmap + transitive allocation of the nested hashmaps + dynamic allocation of the entry.
-                delta_increase = direct_alloc + transitive_alloc + entry_bytes;
+                // services hashmap + transitive allocation of the nested hashmaps.
+                delta_increase = direct_alloc + transitive_alloc;
             }
         }
         if self.config.flush_mode != FlushMode::Periodical {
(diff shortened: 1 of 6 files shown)


threescale/src/utils.rs
```diff
@@ -75,3 +80,42 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
...
+    fn allocated_size(&self) -> usize {
+        self.capacity()
+    }
+}
+
+impl AllocatedSize for AppIdentifier {
+    fn allocated_size(&self) -> usize {
+        match self {
+            AppIdentifier::AppId(app_id, None) => {
+                std::mem::size_of::<AppIdentifier>() + app_id.as_ref().to_string().capacity()
```
unleashed: Note that you are cloning a string to note down how much capacity it needs in the heap. This isn't realistically going to be done at scale, so you'd have to estimate the capacity in the heap based off the string's length.

unleashed (commented):
Hi @NomadXD - I've added a few comments, however I'm not convinced on the costs for this estimation.

Regarding the way the estimation is done, it looks to me as if some things are counted twice because they are on the heap. For example, this is from a very quick glance, but it seems to me you already include the _contents_ of strings in the maps as part of the map's allocated size, yet it seems you count them again separately? This looks confusing to me, as If I was estimating how much a map's allocated size is, I would only take into account the size of the lookup table, not heap allocations that aren't part of the map even if they are referenced by the keys or values. That is, two maps with type `HashMap<String, u64>` with 10 entries each have the same allocated size, regardless of the strings being 10-letter words or if the strings contain each a copy of The Iliad: the contents reside elsewhere in the heap, so not part of the map table itself.

The part that worries me is having to traverse hashmaps to estimate their size. This estimation could be kept at insertion and removal time, but I don't think it's worth it. While the keys can vary, they do only so much, and even when you have hashmaps with variable number of entries, you should still be able to provide a good enough estimation for our purposes based solely on the number of entries without needing to iterate over them (or cloning them to estimate their allocation size).

IOW, I think there is a simpler solution based on how many entries each map has and summing them up to provide an inexact but good enough model on memory consumption that is useful from the PoV of the user. The right thing might just be to let the user specify how many entries each kind of map should have before considering eviction/flushing data because of memory pressure - it can likely work just as well as some sort of memory estimation without the added costs. You don't even need to be super exact about the sizes either, but instead you can figure that if several such maps are over those limits, then perhaps it's already time to start flushing.

singleton-service/src/service/deltas.rs
```diff
@@ -22,7 +25,7 @@ pub struct DeltaStore {
     // Represents the request count. Gets incremented by 1 for each request passing
     // through the proxy. Used together with the capacity attribute to implement a
     // container filling mechansim for cache flush in case of high network traffic scenarios.
-    pub request_count: u64,
+    pub memory_allocated: usize,
```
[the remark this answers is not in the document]
NomadXD: opz. will fix

NomadXD:
@unleashed I added heap allocations just to be close to the real values. But yeah calculating the heap allocations by traversing is not much worth the cost. So I will just consider the allocation for the hashmap keys. (It won't require traversing just multiplication by the number of items). Also I think the rest of the comments are not valid anymore because you have commented in a previous outdated commit.

threescale/src/utils.rs
```diff
@@ -75,3 +80,42 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
         }
     }
 }
+
+impl AllocatedSize for u64 {
+    fn allocated_size(&self) -> usize {
+        8
+    }
+}
```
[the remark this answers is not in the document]
NomadXD: @unleashed this is not there in the latest commit

threescale/src/utils.rs
```diff
@@ -75,3 +80,42 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
...
+            }
+        }
+    }
+}
+
+impl AllocatedSize for HashMap<String, u64> {
+    fn allocated_size(&self) -> usize {
+        let element_size: usize = std::mem::size_of::<String>() + std::mem::size_of::<u64>();
+        let allocated = self.capacity() * element_size;
+        let occupied: usize = self.iter().map(|(key, _)| key.allocated_size()).sum();
```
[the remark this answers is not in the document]
NomadXD: Yeah will consider only the hashmap allocations.

threescale/src/utils.rs
```diff
@@ -75,3 +80,42 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
...
+                std::mem::size_of::<AppIdentifier>() + user_key.as_ref().to_string().capacity()
+            }
+        }
+    }
+}
+
+impl AllocatedSize for HashMap<String, u64> {
+    fn allocated_size(&self) -> usize {
+        let element_size: usize = std::mem::size_of::<String>() + std::mem::size_of::<u64>();
+        let allocated = self.capacity() * element_size;
```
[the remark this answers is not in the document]
NomadXD: The other option is to use `self.len()` which is an underestimate.

threescale/src/utils.rs
```diff
@@ -75,3 +80,42 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
...
+            }
+            AppIdentifier::AppId(app_id, Some(app_key)) => {
+                std::mem::size_of::<AppIdentifier>()
+                    + app_id.as_ref().to_string().capacity()
+                    + app_key.as_ref().to_string().capacity()
+            }
+            AppIdentifier::UserKey(user_key) => {
+                std::mem::size_of::<AppIdentifier>() + user_key.as_ref().to_string().capacity()
+            }
+        }
```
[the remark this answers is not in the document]
NomadXD: outdated

threescale/src/utils.rs
```diff
@@ -75,3 +83,36 @@ pub fn add_hierarchy_to_metrics(hierarchy: &Hierarchy, metrics: &mut Metrics) {
...
+impl AllocatedSize for AppIdentifier {
+    // Returns the dynamically allocated value of a AppIdentifier on the heap.
+    fn dynamic_allocated_size(&self) -> usize {
+        match self {
+            AppIdentifier::AppId(app_id, None) => app_id.as_ref().to_string().capacity(),
+            AppIdentifier::AppId(app_id, Some(app_key)) => {
+                app_id.as_ref().to_string().capacity() + app_key.as_ref().to_string().capacity()
+            }
+            AppIdentifier::UserKey(user_key) => user_key.as_ref().to_string().capacity(),
+        }
```
rahulanand16nov, line 105: Since AppId and UserKey are structs wrapping `String`, can we get **its** capacity? Here we are converting `&str` to `String` and getting its capacity. If not, how about just counting the number of characters and we know each char = 1 byte.

rahulanand16nov: I am not sure if it will work but implementing AllocatedSize for AppId and UserKey is worth a try.

rahulanand16nov (approved):
Just check my comments before merging else looks good to me.

threescale/src/utils.rs
rahulanand16nov: Looks like @unleashed mentioned the same things above. So, you can ignore my comments.

NomadXD:
@unleashed As you suggested , I removed all the dynamic allocations. So only the allocations for the hashmap keys and values are considered. 

1. If a service is added, it will increase the services hashmap with 1 entry, applications hashmap with 1 entry and metrics hashmap with 1 or more entries. 
2. If a application is added to the applications hashmap of a particular service, then application hashmap will get increased by 1 entry, metrics hashmap will get increased by 1 or more entries. 
3. If a new metric is added to the metrics hashmap, it will increase the metrics hashmap by 1 entry.

So in all the cases direct allocation and transitive allocation of hashmap keys and values is considered without dynamic allocation on the heap. Also I changed the `COUNT * self.capacity()` to `COUNT * self.len()` for hashmap in order to calculate only for the occupied instead of allocated.

threescale/src/utils.rs
NomadXD: @rahulanand16nov This is no longer required as suggested by @unleashed . Dynamic allocations are not considered.

singleton-service/src/service/deltas.rs
```diff
@@ -19,10 +20,16 @@ pub struct DeltaStore {
...
-    // through the proxy. Used together with the capacity attribute to implement a
-    // container filling mechansim for cache flush in case of high network traffic scenarios.
-    pub request_count: u64,
+    // Represents a value which is proportional to the memory allocation (underestimate appoximation).
+    // Only the memory allocation for deltas hashmap keys and values are considered. Used together with
+    // the capacity attribute to implement a container filling mechansim for cache flush in case of high
+    // network traffic scenarios.
+    // NOTE : Do not interpret this values as the memory allocation of the delta store because only the
+    // allocation of hashmap keys and values are considerd. Dynamic memory allocation and other control
+    // bytes are not considerd here. So not suitable to take decisions for memory management.
```
unleashed, line 29: typo `considered`

unleashed (approved):
Ok, the only additional point which is interesting to note down is that when collisions happen when trying to insert an entry in a hash map, many algorithms can be used which might or might not increase the memory usage (ie. such as allocating a linked list). That said I don't think you can have access to those internals and it would probably not be worth it anyway to try to account for it.

PR #33, opened 2021-07-16
Upstream info from previous filter

rahulanand16nov:
This PR now makes cache use of upstream info from the previous filter in the chain and passes the same into the MQ to be used by the singleton.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,7 @@ dependencies = [
  "thiserror",
  "threescale",
  "threescalers",
+ "url",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


unleashed (commented):
LGTM

NomadXD (changes requested):
@rahulanand16nov You have modified the source files of the auth sub module here. If you need to change anything there , send a PR to its upstream. Don't make changes to that here since those won't be reflected in its upstream.

rahulanand16nov:
> @rahulanand16nov You have modified the source files of the auth sub module here. If you need to change anything there , send a PR to its upstream. Don't make changes to that here since those won't be reflected in its upstream.

I just updated the submodule using `git pull` as it was pointing to an older commit.

NomadXD (approved):
Oh sorry didn't notice that. You can do remote update via  `--remote --merge` flags. Foreach approach also does the same thing. LGTM.

PR #36, opened 2021-07-21
Add job for build and clippy with no feature included 

rahulanand16nov:
Since now we have features that can be removed, we need to run CI checks when no feature is allowed to be built into the modules.

### DIFF
--- a/.github/workflows/clippy.yaml
+++ b/.github/workflows/clippy.yaml
@@ -28,3 +28,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-targets --all-features -- -D warnings
+
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-targets -- -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/clippy.yaml
```diff
@@ -28,3 +28,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-targets --all-features -- -D warnings
+
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-targets -- -D warnings
```
NomadXD, line 35: Shouldn't this be `--all-targets -- -D warnings` without the `--` in the middle ?? Also this is added as a step of the same job. So it will execute after the previous with `--all-features`. So it doesn't make sense because this is like a subset of the previous. Not sure whether that's the behavior you want.

rahulanand16nov: `--all-features` means include every feature that is present in the code; without it, only default features will be present. You can even look at the time difference between these two in the CI.

NomadXD (approved):

PR #37, opened 2021-07-22
Easing integration tests using a backend state manager

rahulanand16nov:
This PR adds a struct to easily manager backend (Apisonator) state with functions:
- `Pop()`: to pop the last state added.
- `Flush()`: to clear off every state added till then. No need to write Delete* functions anymore 😄 
- `Push()`: to add a new state.

Plus, did some mopping in the cache side as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ clean-apisonator:
 
 auth: ## Build threescale_wasm_auth filter. 
 	@echo "> Building threescale_wasm_auth filter"
-	git submodule foreach git pull origin main
+	git submodule update --init
 	cd threescale-wasm-auth && \
 	make clean && \
 	make build
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ clean-apisonator:
 
 auth: ## Build threescale_wasm_auth filter. 
 	@echo "> Building threescale_wasm_auth filter"
-	git submodule update --init
+	git submodule foreach git pull origin main
 	cd threescale-wasm-auth && \
 	make clean && \
 	make build
--- a/integration-tests/app_id_test.go
+++ b/integration-tests/app_id_test.go
@@ -12,6 +12,7 @@ import (
 
 type AppCredentialTestSuite struct {
 	suite.Suite
+	backend      Backend
 	ServiceID    string
 	ServiceToken string
 	AppID        string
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ clean-apisonator:
 
 auth: ## Build threescale_wasm_auth filter. 
 	@echo "> Building threescale_wasm_auth filter"
-	git submodule foreach git pull origin main
+	git submodule update --init
```
NomadXD, line 58: We need to pull to clarify that we are in sync with the auth upstream. If we are behind, then the CI will fail saying unrelated git history so we know that we are behind. Otherwise we'll be running the CI with an outdated version of the auth.

integration-tests/app_id_test.go
```diff
@@ -165,17 +151,17 @@ func (suite *AppCredentialTestSuite) TestUnlimitedUserKey() {
...
+	if err := suite.backend.Pop(); err != nil {
 		suite.Errorf(err, "Failed to delete applications: %v")
 	}
 }
 
 func (suite *AppCredentialTestSuite) TestUnlimitedAppId() {
 	// Add a new unlimited app
-	if err := AddApplication(suite.ServiceID, "unlimited_app_id", suite.PlanID); err != nil {
+	if err := suite.backend.Push("app", []interface{}{suite.ServiceID, "unlimited_app_id", suite.PlanID}); err != nil {
 		suite.Errorf(err, "Error adding an application: %v")
```
NomadXD, line 165: `suite.Errorf` is just a wrapper around the error. Use `require.Nilf` or `assert.Nilf` to fail the test in case of an error. `require` stop and fails the test right after the error and `assert` fails the test after executing the rest of the test. In this case `assert` suites more but if one of the deletions failed, then we will have issues with the successor test cases because the apisonator already contains apps, services etc from the previous test cases. To overcome this issue we can randomly generate the app_ids, service_ids from the code itself. for now just adding a assert or require seems okay.

NomadXD (changes requested):

Makefile
rahulanand16nov: I am aware of the fact that wasm-auth module can be behind compared to upstream since I added the `foreach...` statement because of the same exact issue. But if you look at commit `e25e97b`, you'll see that it doesn't know how to manage conflict and it can create a merge commit if we force it to pull. That's why I think it's best to manually update the submodule to avoid critical changes by the code, on which we have no control.

integration-tests/app_id_test.go
rahulanand16nov: Sure, I'll change this and yes, I am thinking of adding a new PR allowing changes mentioned by you.

Makefile
NomadXD: we shouldn't force(by default its not forcing unless we pass `--force`). it won't merge if it sees unrelated git histories. the build will fail and then we know we are behind. then we can update manually. otherwise we won't know.

integration-tests/app_id_test.go
NomadXD: either will work. only difference is assert will delete the rest of the resources.

Makefile
rahulanand16nov: Build won't fail but actually, the integration job may fail which can be tested using `make integration` before making a PR.

NomadXD: No in this case build fails. For building the auth git submodule update is required and it will fail if it sees the unrelated git history. Your CI for this PR failed because of that. 
https://github.com/3scale-labs/gsoc-wasm-filters/actions/runs/1056006285

Both ways the sub module update is done manually. Only difference is if we pull we know that the sub module is lagging behind. If we don't pull then someone has to always look whether the auth is updated manually which is not very maintainable.

rahulanand16nov: You'll only know that submodule is lagging behind only when there is an unrelated git history else it will pull and use the latest version even though you have a previous commit for the sub-module in this repo. So, someone has to be always aware of that lag.

NomadXD: If someone updates the auth upstream and we have a previous version, then we'll always get a unrelated history.

rahulanand16nov: Updated.

NomadXD (approved):
LGTM

PR #38, opened 2021-07-22
Singleton flush test and few fixes

NomadXD:
This PR includes
- Adding apisonator worker
- Singleton flush test
- Removing upstream struct and custom serializer


unleashed (commented):
LGTM, but let's have @rahulanand16nov take a look too.

integration-tests/singleton_flush_test.go
```diff
@@ -0,0 +1,237 @@
...
+		}, 4*time.Second, 500*time.Millisecond, "Error deleting the service")
+
+		for _, app := range suite.AppIDs {
+			require.Eventually(suite.T(), func() bool {
+				err := DeleteApplication(suite.ServiceID, app)
+				if err != nil {
+					return false
+				}
+				return true
+			}, 4*time.Second, 500*time.Millisecond, "Error deleting the app")
```
rahulanand16nov, line 216: How about `suite.states.Flush()`?

rahulanand16nov (changes requested):
Mostly LTGM! Just deleting state can be done using `Flush()` and `Pop()`.

integration-tests/singleton_flush_test.go
```diff
@@ -0,0 +1,237 @@
...
+		configErr := configureSingletonFlush("ContainerLimit")
+		require.Nilf(suite.T(), configErr, "Error configuring envoy.yaml for container flush : %v", configErr)
+
+		require.Eventually(suite.T(), func() bool {
+			fmt.Printf("Creating service with service_id: %s, service_token:%s", suite.ServiceID, suite.ServiceToken)
+			serviceErr := CreateService(suite.ServiceID, suite.ServiceToken)
+			if serviceErr != nil {
+				return false
+			}
+			return true
```
rahulanand16nov, line 54: How about `Push()`?

integration-tests/singleton_flush_test.go
```diff
@@ -0,0 +1,237 @@
...
+	var singletonConfig = fmt.Sprintf(`
+      name: envoy.bootstrap.wasm
+      typed_config:
+        '@type': type.googleapis.com/envoy.extensions.wasm.v3.WasmService
+        singleton: true
+        config:
+          name: "singleton_service"
+          root_id: "singleton_service"
+          configuration: 
+            "@type": type.googleapis.com/google.protobuf.StringValue
```
rahulanand16nov, line 105: So much code just to change 1 value. 🤔

integration-tests/singleton_flush_test.go
NomadXD: Yeah will change to those in the next PR with the other test cases for the singleton.

integration-tests/singleton_flush_test.go
NomadXD: Yeah will change to those in the next PR with the other test cases for the singleton.

rahulanand16nov (approved):

integration-tests/singleton_flush_test.go
NomadXD: Yeah this is how you usually do yaml formatting in golang. With the golang limitation of unmarshalling, you can either define a struct with the related tags or dynamically handle with interfaces. For very large yaml, since it's difficult to define a struct, dynamic handling is done. But for deep nested, to prevent the nested reads with interfaces, whole section is formatted.

PR #40, opened 2021-07-27
Singleton integration tests 

NomadXD:
This PR includes
- Integration tests for singelton flush based on periodical flush
- Added UUID for keys to prevent dependency between test cases in case helper methods fail (If we use the same hard coded service ids , if the clean up methods fail for some reason then the successor tests also will fail when trying to create a service which already exists)
- Few other minor fixes

### DIFF
--- a/integration-tests/app_id_test.go
+++ b/integration-tests/app_id_test.go
@@ -3,8 +3,11 @@ package main
 import (
 	"fmt"
 	"net/http"
+	"os"
 	"testing"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
@@ -23,15 +26,13 @@ type AppCredentialTestSuite struct {
 }
 
 func (suite *AppCredentialTestSuite) SetupSuite() {
-	err := StartProxy("./", "./envoy.yaml")
-	require.Nilf(suite.T(), err, "Error starting proxy: %v", err)
 	// Initializing 3scale backend state
 	suite.AppID = "test-app-id"
 	suite.AppKey = "test-app-key"
 	suite.UserKey = "test-user-key"
-	suite.ServiceID = "test-service-id"
+	suite.ServiceID = uuid.NewString()
 	suite.PlanID = "test-plan-id"
-	suite.ServiceToken = "test-service-token"
+	suite.ServiceToken = uuid.NewString()
 	suite.metrics = []Metric{
 		{"hits", "1", []UsageLimit{
 			{Day, 10000},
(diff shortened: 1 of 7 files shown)


integration-tests/app_id_test.go
```diff
@@ -23,15 +26,13 @@ type AppCredentialTestSuite struct {
...
-	require.Nilf(suite.T(), err, "Error starting proxy: %v", err)
 	// Initializing 3scale backend state
 	suite.AppID = "test-app-id"
 	suite.AppKey = "test-app-key"
 	suite.UserKey = "test-user-key"
-	suite.ServiceID = "test-service-id"
+	suite.ServiceID = uuid.NewString()
 	suite.PlanID = "test-plan-id"
-	suite.ServiceToken = "test-service-token"
+	suite.ServiceToken = uuid.NewString()
```
rahulanand16nov, line 35: Since we have serialized tests, I don't think so we should have a case where ITs fail due to duplicate service without any error on the writer side. I was thinking about parallelizing the tests but that would require a different solution than this, which I can discuss if you want.

integration-tests/app_id_test.go
```diff
@@ -40,23 +41,66 @@ func (suite *AppCredentialTestSuite) SetupSuite() {
...
-	usageErr := suite.backend.Push("usages", []interface{}{suite.ServiceID, suite.PlanID, &suite.metrics})
-	require.Nilf(suite.T(), usageErr, "Error: %v", usageErr)
+	require.Eventually(suite.T(), func() bool {
+		serviceErr := suite.backend.Push("service", []interface{}{suite.ServiceID, suite.ServiceToken})
+		if serviceErr != nil {
+			fmt.Printf("Error creating the service: %v", serviceErr)
+			return false
+		}
+		return true
+	}, 5*time.Second, 1*time.Second)
```
rahulanand16nov, line 52: I think eventually serves no purpose here since `Push` defers till the response is received and adding a 500ms wait before `Push` returns (I've added that already in my new PR) kind of guarantees that things are processed by the Apisonator.

integration-tests/config_template.yaml
```diff
@@ -100,8 +100,8 @@ static_resources:
                       },
                       "services": [
                         {
-                          "id": {{or .ServiceID "\"test-service-id\""}},
-                          "token": {{or .ServiceToken "\"test-service-token\""}},
+                          "id": "{{or .ServiceID "\"test-service-id\""}}",
+                          "token": "{{or .ServiceToken "\"test-service-token\""}}",
```
rahulanand16nov, line 104: Note here that default case will return `""test-service-id""` and `""test-service-token""`

integration-tests/singleton_flush_test.go
```diff
@@ -38,81 +40,48 @@ func (suite *SingletonFlushTestSuite) SetupSuite() {
...
+// Gets triggered before each test and runs the pre-conditions like setting up apisonator services
+// and generating envoy config from a template.
 func (suite *SingletonFlushTestSuite) BeforeTest(suiteName, testName string) {
 	if testName == "TestSingletonContainerFlush" {
-		configErr := configureSingletonFlush("ContainerLimit")
+		suite.ServiceID = uuid.NewString()
+		suite.ServiceToken = uuid.NewString()
+
+		// Configure the envoy template with ContainerLimit flush.
+		configErr := configureSingletonFlush("ContainerLimit", "100", "60", suite.ServiceID, suite.ServiceToken)
```
rahulanand16nov, line 51: It's just preference but I think if you are using `if testName == ...` then better add these conditions into the test as it is because now, readers have to be aware of two blocks for a single test i.e `BeforeTest` and the test function itself.

rahulanand16nov (changes requested):
Please check the comments provided. Rest, LTGM.

integration-tests/app_id_test.go
NomadXD: There can be issues in the network or in the server side as well.  Also there's another issue with stats. When you create a usage and delete it the stats won't get deleted. We have to delete the stats which we don't do at the moment. So when you create a new service/app combination, it already contains stats. Managing apisonator apps, services and keys is a useless effort since it doesn't provide us any benefit and all the apisonator states get deleted after the tests. We don't have to worry about the apisonator apps, services when we use UUIDs.

integration-tests/app_id_test.go
NomadXD: For things like timeouts and waits, it's better to use `require.Eventually` rather than adding `time.Sleep()` since the test suite already has this feature. It will perform retries periodically according to the way we define the related times. I'm thinking to remove the static `time.Sleep` in proxy start up method and do a check there.

integration-tests/config_template.yaml
NomadXD: okay will fix. I think we don't need a default since we use UUIDs

integration-tests/app_id_test.go
rahulanand16nov: It wait for the function to return success but `cmd.Run` already returns success even though the proxy is still loading so eventually serves no purpose there as well since we are waiting for the proxy to get ready.

integration-tests/app_id_test.go
rahulanand16nov: Weird, I've only seen running tests in parallel but that may be just my experience.

integration-tests/singleton_flush_test.go
NomadXD: okay will add the pre-conditions to the test itself.

integration-tests/app_id_test.go
NomadXD: How you design and write tests depends on the requirement. So I'm not saying that parallel tests are bad. I'm saying that for our case it is not feasible to run several proxy instances for the integration tests. apisonator records the usages as stats. You can see the redis database for the stats. We are not deleting the stats. We don't need to consider those stats since we use UUIDs. Otherwise if you create the new service app key combination, it will have the usage from the previous tests.

integration-tests/app_id_test.go
NomadXD: yeah that's why we need to do a check using eventually to test whether the proxy is up or not.

rahulanand16nov: Can you explain to me how it does that?

NomadXD: From the admin  check for the listener. 
```go
require.Eventually(t, func() bool {
    res, err := http.Get(fmt.Sprintf("http://localhost:%d/listeners", adminPort))
    if err != nil {
	    return false
    }
    defer res.Body.Close()
    return res.StatusCode == http.StatusOK
}, 5*time.Second, 100*time.Millisecond, "Envoy has not started")
```

rahulanand16nov: Better solution! Please add this into the PR itself.

NomadXD: No need to have a event to check apisonator. Helper methods returns client errors, server errors and network errors and `Push` returns those errors directly to the caller. In case of such issues, eventually performs retries.

rahulanand16nov (approved):

PR #41, opened 2021-07-27
Bad configuration tests

rahulanand16nov:
This PR adds test cases for when there is a minor mistake in the config file and checks that the proxy does the intended behavior for these cases.

### DIFF
--- a/integration-tests/apisonator.go
+++ b/integration-tests/apisonator.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"time"
 )
 
 // BackendState represents the Apisonator internal state
--- a/integration-tests/apisonator.go
+++ b/integration-tests/apisonator.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 )
 
 // BackendState represents the Apisonator internal state
--- a/integration-tests/main.go
+++ b/integration-tests/main.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"regexp"
 	"text/template"
+	"time"
 )
 
 func main() {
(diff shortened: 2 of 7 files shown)


integration-tests/apisonator.go
```diff
@@ -376,6 +377,8 @@ func executeHTTPRequest(method string, url string, data *[]byte) (*http.Response
 func (backend *Backend) Push(stateName string, params []interface{}) error {
 	defer func() {
```
NomadXD, line 378: There is an issue with the use of `defer` here. Defer executes after the surrounding function returns. So here if you do this way, even when the functions in the switch case returns an error , you are adding a new state to the backend states. I fixed this issue in my PR and should be file after you do a rebase.

integration-tests/apisonator.go
```diff
@@ -404,6 +407,8 @@ func (backend *Backend) Pop() error {
 
 		defer func() {
```
NomadXD, line 408: same issue with the `defer` here.

integration-tests/main.go
```diff
@@ -34,6 +35,7 @@ func GenerateConfig(name string, jsonData []byte) error {
 	if writeErr := ioutil.WriteFile(name, out.Bytes(), 0777); writeErr != nil {
 		fmt.Printf("Error writing temp config file: %v", writeErr)
 	}
+	time.Sleep(100 * time.Millisecond)
```
NomadXD, line 38: Why we need a timeout here ???

NomadXD (approved):
LGTM , check comments

integration-tests/apisonator.go
rahulanand16nov: I was thinking of using named return value to tackle this problem but your solution is also good.

integration-tests/apisonator.go
rahulanand16nov: I think you have fixed it already :D

integration-tests/main.go
rahulanand16nov: I was having issues while testing but forget to remove it before the PR. Good find!

PR #43, opened 2021-08-02
Memory limit on shared data

rahulanand16nov:
This PR adds the ability to limit memory consumed by shared data since it can fill up very quickly in a production environment. What should happen if the memory limit is reached is still undefined and a possible solution is to allow singleton to evict cache as someone needs to store keys present in the cache with some policy for eviction.

### DIFF
--- a/cache-filter/src/configuration.rs
+++ b/cache-filter/src/configuration.rs
@@ -7,6 +7,8 @@ pub struct FilterConfig {
     pub failure_mode_deny: bool,
     /// Number of retries for setting data to cache
     pub max_tries: u32,
+    /// Max memory in bytes that shared data is allowed to use.
+    pub max_shared_memory_bytes: u64,
 }
 
 impl Default for FilterConfig {
@@ -14,6 +16,7 @@ impl Default for FilterConfig {
         FilterConfig {
             failure_mode_deny: true,
             max_tries: 5,
+            max_shared_memory_bytes: 4294967296, // equivalent to 4GB
         }
     }
 }
--- a/cache-filter/src/configuration.rs
+++ b/cache-filter/src/configuration.rs
@@ -11,12 +11,14 @@ pub struct FilterConfig {
     pub max_shared_memory_bytes: u64,
 }
 
+const DEFAULT_MAX_SHARED_MEMORY: u64 = 4294967296;
+
 impl Default for FilterConfig {
     fn default() -> Self {
         FilterConfig {
             failure_mode_deny: true,
             max_tries: 5,
-            max_shared_memory_bytes: 4294967296, // equivalent to 4GB
+            max_shared_memory_bytes: DEFAULT_MAX_SHARED_MEMORY, // equivalent to 4GB
         }
     }
 }
--- a/threescale/src/proxy.rs
+++ b/threescale/src/proxy.rs
@@ -19,6 +19,8 @@ pub enum CacheError {
     ProxyStatus(u8),
     #[error("deserializing bincode data failed")]
     DeserializeFail(#[from] bincode::ErrorKind),
+    #[error("serializing into bincode format failed")]
+    SerializeFail,
 }
 
 #[derive(Debug, Clone, Eq)]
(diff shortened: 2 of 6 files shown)


cache-filter/src/configuration.rs
```diff
@@ -7,13 +7,16 @@ pub struct FilterConfig {
...
+    /// Max memory in bytes that shared data is allowed to use.
+    pub max_shared_memory_bytes: u64,
 }
 
 impl Default for FilterConfig {
     fn default() -> Self {
         FilterConfig {
             failure_mode_deny: true,
             max_tries: 5,
+            max_shared_memory_bytes: 4294967296, // equivalent to 4GB
```
unleashed, line 19: Never hardcode these, always use a constant defined in a prominent place.

threescale/src/proxy.rs
```diff
@@ -92,15 +96,79 @@ pub fn set_app_id_to_cache(user_key: &UserKey, app_id: &AppId) -> Result<(), Cac
...
-        key,
-        Some(&bincode::serialize::<Application>(app).unwrap()),
-        Some(cas),
-    ) {
-        Ok(()) => true,
-        Err(e) => {
-            debug!("set operation failed for key: {} : {:?}", key, e);
-            false
+    let prev_memory_usage = get_cache_pair_size(key) as i32;
+    let serialized_app = bincode::serialize::<Application>(app).unwrap();
```
unleashed, line 100: What would we need to do if this failed to serialize? Log something and return `false`?

threescale/src/proxy.rs
```diff
@@ -92,15 +96,79 @@ pub fn set_app_id_to_cache(user_key: &UserKey, app_id: &AppId) -> Result<(), Cac
...
+            // increasing since we have no option of deleting the data. So, check your calcs!
+            final_size = SHARED_MEMORY_INITIAL_SIZE;
+        } else {
+            final_size = current_size.saturating_sub((-delta).try_into().unwrap());
+        }
+    } else {
+        final_size = current_size.saturating_add((-delta).try_into().unwrap());
+    }
+
+    if let Err(e) = set_shared_data(SHARED_MEMORY_KEY, Some(&final_size.to_be_bytes()), None) {
```
unleashed: You'll probably want to get the `cas` from `get_shared_data` above and reuse it here - if the call failed _with a CasMismatch_ then someone else updated this before you did, so... probably you don't need to care about it and can assume it was properly updated?

threescale/src/proxy.rs
```diff
@@ -92,15 +96,79 @@ pub fn set_app_id_to_cache(user_key: &UserKey, app_id: &AppId) -> Result<(), Cac
...
     }
+    true
+}
+
+// returns memory used in bytes for both key and value pair stored.
+fn get_cache_pair_size(key: &str) -> usize {
+    match get_shared_data(key) {
+        Ok((Some(bytes), _)) => key.len() + bytes.len(),
+        Ok((None, _)) => key.len(),
+        Err(_) => 0,
```
unleashed, line 120: If this call fails then you'll get back a 0, but the code will happily try to update the key anyway - this needs error handling to bail out early before trying to write incorrect data (which likely won't work if `get_shared_data` fails, but who knows what could be happening in the host side - ie. bugs, cpu-on-fire condition, etc).

threescale/src/proxy.rs
```diff
@@ -92,15 +96,83 @@ pub fn set_app_id_to_cache(user_key: &UserKey, app_id: &AppId) -> Result<(), Cac
...
+            final_size = memory_used.saturating_sub((-delta).try_into().unwrap());
+        }
+    } else {
+        final_size = memory_used.saturating_add(delta.try_into().unwrap());
+    }
+
+    if let Err(e) = set_shared_data(
+        SHARED_MEMORY_COUNTER_KEY,
+        Some(&final_size.to_be_bytes()),
+        None,
```
unleashed, line 170: Check out https://github.com/3scale-labs/gsoc-wasm-filters/pull/43#discussion_r680867751

cache-filter/src/configuration.rs
rahulanand16nov: constant added

threescale/src/proxy.rs
rahulanand16nov: Handled

threescale/src/proxy.rs
rahulanand16nov: Used CAS from above.

threescale/src/proxy.rs
rahulanand16nov: Added better error handling.

threescale/src/proxy.rs
rahulanand16nov: Fixed

NomadXD (approved):
LGTM

PR #44, opened 2021-08-03
Delete apps stored in the cache

NomadXD:
Deleting apps stored in the cache when the 3scale SM API returns an authorization error.


rahulanand16nov (approved):
LGTM.

PR #46, opened 2021-08-07
Unique callouts

rahulanand16nov:
I recently implemented the PoC for avoiding multiple callouts, at the same time, to SM API that correspond to the same application. This PR ports that PoC and improves on it with better error handling and logging. I am thinking of merging this into a separate branch until more testing is done on it and performance improvement can be seen in benchmarks.

There is a very high chance that there are critical problems so please have a close look and let me know ways in which this can be improved.

Thanks!


cache-filter/src/filter/http.rs
```diff
@@ -243,6 +243,48 @@ impl CacheFilter {
...
+
+        if let Err(e) = set_shared_data(&request_key, None, None) {
+            warn!(
+                self.context_id,
+                "failed to delete the callout-lock from shared data: {}: {:?}", request_key, e
+            );
+            return Err(e);
+        }
+        Ok(())
+    }
```
unleashed, line 349: As a test/debugging measure you might want to call first `get_shared_data` so you can check that the lock exists. Unfortunately the current API won't give you this information, so I'd suggest we add this to help find out potential issues (ie. trying to free a non existing lock).

unleashed:
I think the mechanism looks ok, @rahulanand16nov, but best to test extensively.

cache-filter/src/filter/http.rs
rahulanand16nov: Added.

PR #47, opened 2021-08-09
Benchmark results for HTTP latency

NomadXD:
This PR includes,
- Configs for benchmark performed to compare 3 scenarios for HTTP latency.

1.  Auth + cache + singleton
2. Auth filter only
3. Without plugins

- Reports from the benchmark performed.
- Analysis of the benchmark results.


rahulanand16nov (approved):
Looks great!

PR #50, opened 2021-08-17
Add unique callout module

rahulanand16nov:
This PR includes the unique callout module and adds some fields left in the last PR.


NomadXD (approved):
LGTM. Is there any reason for setting the shared data value to `root_id` ?

rahulanand16nov:
> LGTM. Is there any reason for setting the shared data value to `root_id` ?

`root_id` represents which thread id and can be used to identify it by whosoever is freeing the lock.

cache-filter/src/unique_callout.rs
```diff
@@ -0,0 +1,129 @@
+use crate::{info, warn};
+use proxy_wasm::{
+    hostcalls::{get_shared_data, set_shared_data},
+    types::Status,
+};
+use threescale::proxy::CacheKey;
+
+// Callout lock is acquired by placing a key-value pair inside shared data.
+// Since, only one thread is allowed to access shared data (host uses mutex) thus only one winner;
```
unleashed, line 9: Since this is far from trivial, this needs a TL;DR about the algorithm regardless of comments inside the function. Also remember that this is based on what the Envoy proxy is doing, since the spec does not go into details - so make sure to document these assumptions.

cache-filter/src/unique_callout.rs
```diff
@@ -0,0 +1,129 @@
...
+        context_id,
+        "thread(id: {}): trying to set callout-lock for request(key: {})", root_id, request_key
+    );
+
+    // check if lock is already acquired or not
+    match get_shared_data(&request_key)? {
+        (_, None) => {
+            // lock not acquired yet, try to set lock.
+            // Note: Since two threads can get None for CAS at the same time, to avoid
+            // overwriting of lock by another thread, use any value for CAS other than 0.
```
unleashed, line 28: Missing info: what's the expected value of CAS on success? The reader does not have details at this point about these implicit behavior.

cache-filter/src/unique_callout.rs
```diff
@@ -0,0 +1,129 @@
...
+            info!(
+                context_id,
+                "thread(id: {}): callout-lock for request(key: {}) already acquired by thread(id: {})",
+                root_id,
+                request_key,
+                lock_acquired_by
+            );
+            Ok(false)
+        }
+        (None, Some(cas)) => {
```
unleashed, line 63: Missing comments: When does this case happen? How can someone have a key with no value inside? ie. link this to how the lock is freed.

cache-filter/src/unique_callout.rs
```diff
@@ -0,0 +1,129 @@
...
+                    Ok(false)
+                }
+                Err(e) => Err(e),
+            }
+        }
+    }
+}
+
+// NOTE: Right now, there is no option of deleting the pair instead only the value can be erased,
+// and it requires changes in the ABI so change this because it will lead to better memory usage.
```
unleashed, line 96: One important issue here is whether the Proxy-WASM specification really goes into detail around this. Unfortunately it is very far from giving these details, so we need to work with the implementation we have in Envoy - always note these details to clearly separate the concrete Envoy implementation which is part of our assumptions from the spec, which might not define at all how these details interact with our client code.

unleashed (commented):
Looks good, but there are a few missing things. Check my comments, but essentially we need to document this better in inline comments to let the reader better understand our assumptions and the way we exploit the Envoy implementation, not quite so the spec.

Additionally, there is no mention to the possibility of a random match of the CAS where two threads could get a hold of the lock - this needs documenting and explaining, because people will assume that it never happens, and in a system running long enough the chances grow exponentially, with `long enough` being such an indeterminate period of time that it could happen within a relatively short amount of time in a loaded enough proxy.

As a follow up to this last consideration: any thoughts on mitigation - if so (or if you can't think of any), also add them in a suitable place here, so that the next person knows the issues and solutions you considered.

cache-filter/src/unique_callout.rs
rahulanand16nov: Added a huge comment above explaining how setting and freeing is done for the lock.

cache-filter/src/unique_callout.rs
rahulanand16nov: I think logs also act as a way to build context around what is happening, right?

cache-filter/src/unique_callout.rs
rahulanand16nov: Same as above.

cache-filter/src/unique_callout.rs
rahulanand16nov: I think adding a big comment with granular details helps here as well but if you want to re-iterate some SDK info here then I can do that.

rahulanand16nov:
Now I had the time to rethink the case when multiple callouts happen, its probability is actually way lower than 1/{u32::MAX} because you need a very specific condition to happen: lock was acquired, freed with CAS=1, then threads try to re-acquire the lock (cache-entry deleted).

PR #53, opened 2021-08-23
Unique Callout using MQ v0.1

rahulanand16nov:
This PR implements the first version of the working unique callout feature using message queues.

Note: There are cases in which resuming a waiting context fails and request stays in the state of limbo (e.g. if auth callout timeouts). These cases will be slowly fixed since PR is already very large and needs review before moving forward.

### DIFF
--- a/cache-filter/src/filter/root.rs
+++ b/cache-filter/src/filter/root.rs
@@ -141,6 +141,7 @@ impl RootContext for CacheFilterRoot {
                                     self.context_id,
                                     "failed to map user_key to app_id cache key pattern: {:?}", e
                                 );
+                                waiters.remove(&context_to_resume);
                                 return;
                             }
                         }
@@ -167,6 +168,7 @@ impl RootContext for CacheFilterRoot {
                             "failed to fetch application from cache: {:?}", e
                         ),
                     }
+                    waiters.remove(&context_to_resume);
                 })
             }
             Ok(None) => warn!(
--- a/cache-filter/src/filter/root.rs
+++ b/cache-filter/src/filter/root.rs
@@ -54,6 +54,13 @@ impl RootContext for CacheFilterRoot {
         self.id = self.rng.next_u32();
         info!(self.context_id, "root initialized with id: {}", self.id);
 
+        // Initializing a thread-specific message queue
+        let queue_id = self.register_shared_queue(&self.id.to_string());
+        info!(
+            self.context_id,
+            "root({}): registered thread-specific MQ ({})", self.id, queue_id
+        );
+
         //Check for the configuration passed by envoy.yaml
         let configuration: Vec<u8> = match self.get_configuration() {
             Some(c) => c,
--- a/cache-filter/src/filter/root.rs
+++ b/cache-filter/src/filter/root.rs
@@ -1,12 +1,18 @@
 use crate::configuration::FilterConfig;
 use crate::filter::http::CacheFilter;
 use crate::rand::thread_rng::{thread_rng_init_fallible, ThreadRng};
+use crate::unique_callout::WAITING_CONTEXTS;
+use crate::utils::request_process_failure;
 use crate::{debug, info, warn};
 use proxy_wasm::{
+    hostcalls::set_effective_context,
     traits::{Context, HttpContext, RootContext},
     types::{ContextType, LogLevel},
 };
-use threescale::{proxy::CacheKey, structs::ThreescaleData};
+use threescale::{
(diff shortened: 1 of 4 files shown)


cache-filter/src/unique_callout.rs
```diff
@@ -129,30 +170,171 @@ pub fn free_callout_lock(
...
+                                    return Err(UniqueCalloutError::SerializeFail {
+                                        what: "Vec<CalloutWaiter>",
+                                        reason: *e,
+                                    })
+                                }
+                            };
+                        if let Err(Status::CasMismatch) =
+                            set_shared_data(&waiters_key, Some(&serialized_cw), Some(cas))
+                        {
+                            continue;
```
NomadXD, line 242: It's nice to add a log here so we know `CasMismatch` happened.

cache-filter/src/unique_callout.rs
```diff
@@ -129,30 +170,171 @@ pub fn free_callout_lock(
...
+                                reason: *e,
+                            })
+                        }
+                    };
+                // Again exploiting host implementation to avoid multi-initialization:
+                let cas = cas.unwrap_or(1);
+                if let Err(Status::CasMismatch) =
+                    set_shared_data(&waiters_key, Some(&serialized_cw), Some(cas))
+                {
+                    continue;
```
NomadXD, line 274: It's nice to add a log here so we know `CasMismatch` happened.

cache-filter/src/unique_callout.rs
```diff
@@ -129,30 +170,171 @@ pub fn free_callout_lock(
...
+            }
+            Err(e) => {
+                return Err(UniqueCalloutError::DeserializeFail {
+                    what: "Vec<CalloutWaiter>",
+                    reason: *e,
+                });
+            }
+        },
+        Ok((None, _)) => {
+            // This can happen either someother thread freed waiting contexts or
```
NomadXD, line 330: `"This can happen either someother thread freed waiting contexts"` - How is this possible ? Response to an external http requests comes to the same thread that made the request. And as I understood with this implementation, `waiters_key` can be acquired by only one thread at a time.

cache-filter/src/filter/root.rs
```diff
@@ -83,6 +96,90 @@ impl RootContext for CacheFilterRoot {
...
+                    Err(e) => {
+                        warn!(
+                            self.context_id,
+                            "failed to parse message({}) from MQ into ctxt id: {}", ctxt_str, e
+                        );
+                        return;
+                    }
+                };
+
+                WAITING_CONTEXTS.with(|refcell| {
```
NomadXD, line 118: After resuming the http request, `context` stored in `WAITING_CONTEXTS` should be removed.

cache-filter/src/unique_callout.rs
rahulanand16nov: Not possible but just in case some changes the implementation by mistake in the future that's why I added that there.

cache-filter/src/filter/root.rs
rahulanand16nov: Fixed

cache-filter/src/unique_callout.rs
rahulanand16nov: Added

cache-filter/src/unique_callout.rs
rahulanand16nov: Added

NomadXD:
I think there's a concurrency issue here as well. For example consider the scenario with threads `A`, `B`, `C` and `D`. Follow the sequence of events.
1. Thread `A` receives a request for the key combination `foo` and performs a http request due to cache miss.
2. Before the response, thread `B` and `C` receives request for the same key `foo` and they get added to the waiting list.
3. Right before the response or right after the response for the http call made in thread `A`, thread `D` receives a request for the same key `foo`. At this point thread `A` owns the callout lock.  So thread `D` sees that and try to add an entry to the waitlist.
4. But before adding an entry to waitlist from thread `D`, thread `A` releases the lock and resume the waiters. Thread `D` adds the entry to the waitlist after and it hangs there since there is no actual pending request.

I think with envoy's non-blocking threading model, this is very unlikely to happen. But I think still there's a possibility specially in high concurrency setups where the number of worker threads > hardware threads in the host.

rahulanand16nov:
@NomadXD That's right! You have described what I said above in a more granular way but there are many more cases in which this type of problem can happen and I have them in my mind and how to tackle them ;)

Good to have it documented here as well since it can help someone who is trying to find more details on the unique callout module.

NomadXD (approved):

PR #58, opened 2021-08-27
Simplifying request failure methods

rahulanand16nov:
This PR simplifies request failure methods to use only 1 ref to CacheFilter object instead of two.

Also, cache hit log, and metric is moved to a better location.


NomadXD (approved):
LGTM

PR #59, opened 2021-08-27
Unique callout benchmarks

NomadXD:
This PR includes the benchmark results, reports and generated diagrams for the unique callouts benchmarks.


rahulanand16nov (approved):
I found the same thing in my manual testing as well and it's good to have concrete data to showcase.

rahulanand16nov:
Looks like you also fixed the same issue but I prefer to make this PR only about "benchmarks" instead of also fixing CI issues to give a better context to the reader. I should have messaged you about what I am working and I thought you were working on seg fault issue. Sorry about that!

PR #60, opened 2021-09-06
Update wasm-auth and tests to fix CI

rahulanand16nov:
Due to the last update to wasm-auth and changes to the link for self-maintained Rust SDK, CI was not working at all and this PR fixes those issues by updating config files and the submodule.

Also, when auth callout timed out, there was no response being sent back to the client, which is also fixed in this PR.

@NomadXD Please merge this before merging #59


rahulanand16nov:
@NomadXD Yeah I know and replied there already. Hate this confusion but when I was clicking on "create pull request" I saw you were still having the issue with CI and created it regardless. Well, still I prefer to have two separate PRs one only for fixing CI (this one) and benchmarks (#59 )

NomadXD (changes requested):
Let's use the header way to send app ids and tokens since it's the usual way of sending tokens and ids.

rahulanand16nov:
> Let's use the header way to send app ids and tokens since it's the usual way of sending tokens and ids.

I don't think there is a "usual" way to sending tokens or ids since it depends on how configs are written, right?

NomadXD:
@rahulanand16nov No the usual way I've seen with the APIs is with the header, not with the query. Refer the RFC standards here https://datatracker.ietf.org/doc/html/rfc7235#section-4.2

rahulanand16nov:
@NomadXD Idk how relevant that link is for us here but aren't we using `authorization` header for app-id pattern in wasm-auth? Of course, we can just redefine config to use `x-app-id/key` header. I can allow both of these patterns in the config if you want?

NomadXD:
@rahulanand16nov It is relevant because those are standards defined in the HTTP spec and we are dealing with HTTP protocol here. We are not using an authorization header here but using what's called an extension header which serves the same purposes as authorization header with a name specific to the organisation.

rahulanand16nov:
@NomadXD Hmmm, interesting! Please have a look again and rebase #60 so we can start tackling more interesting work :)

NomadXD (approved):

PR #61, opened 2021-09-07
Allow unique-callout module to be dis/en-abled using cargo flags

rahulanand16nov:
As previously discussed, even though unique callouts perform better in accuracy and performance, there is still doubt about its reliability. So, it's best to allow users to enable or disable this module using cargo feature flags.

To minimize the use of `cfg` all around the code (it's an anti-pattern to do so), I have added dead code in the `unique_callout.rs` to have them as placeholders if the module is disabled.

After this PR is merged, I'll create a new PR to merge this branch into main to squash all the commits into one merge commit.


rahulanand16nov:
All the CI checks are working fine locally but if someone is still unsure, note that merge PR will run the CI regardless ;)

NomadXD (approved):
LGTM

unleashed (changes requested):
Yes, it's a total antipattern - we don't want to constantly add `#[cfg(feature = "unique_callout")]` everywhere. What we want instead is 1-2 such instances in which one of those is used to include the relevant code in a module, possibly with a `#[cfg(not(feature = "unique_callout"))]` including a dummy module with the same interface that does nothing (other options are possible), and then perhaps, very sparingly and with good justification, use it elsewhere to enable/disable other code.

rahulanand16nov:
@unleashed Added a dummy module with the reasoning behind it as well. I think this is the minimum number of `cfg` attributes that are needed and cannot decrease the count more due to coupling between the `unique_callout` and other modules.

cache-filter/src/filter/http.rs
```diff
@@ -1,9 +1,12 @@
+#[cfg(not(feature = "unique_callout"))]
+use crate::dummy::*;
+#[cfg(feature = "unique_callout")]
+use crate::unique_callout::{
+    free_callout_lock_and_notify_waiters, set_callout_lock, SetCalloutLockStatus, WaiterAction,
+};
```
unleashed, line 6: I think a better solution could be: (not checked, maybe this ends up being not workable)

```rust
#[cfg(not(feature = "unique_callout"))]
use crate::unique_callout_dummy as unique_callout;
#[cfg(feature = "unique_callout")]
use crate::unique_callout;

use unique_callout::{free_callout_lock_and_notify_waiters, set_callout_lock, SetCalloutLockStatus, WaiterAction};
```

cache-filter/src/filter/root.rs
```diff
@@ -1,18 +1,22 @@
...
+// Following imports are only needed when unqiue-callout feature is enabled.
+#[cfg(feature = "unique_callout")]
+use {
+    crate::unique_callout::{WaiterAction, WAITING_CONTEXTS},
+    crate::utils::request_process_failure,
+    proxy_wasm::hostcalls::set_effective_context,
+    threescale::{
+        proxy::{get_app_id_from_cache, get_application_from_cache},
+        structs::AppIdentifier,
+    },
```
unleashed, line 19: If this is only needed for `on_queue_ready`, then we should `use` them only within the body of `on_queue_ready`.

cache-filter/src/lib.rs
```diff
@@ -3,8 +3,11 @@
 const VM_ID: &str = "my_vm_id";
 
 mod configuration;
+#[cfg(not(feature = "unique_callout"))]
+mod dummy;
 mod filter;
 mod log;
 mod rand;
+#[cfg(feature = "unique_callout")]
```
unleashed, line 11: Renaming `dummy` to `unique_callout_dummy` will help place these two together.

cache-filter/src/unique_callout.rs
```diff
@@ -60,6 +60,7 @@ pub enum WaiterAction {
 }
 
 // This enum is used to give out status for an http context trying to get callout-lock.
+#[cfg(feature = "unique_callout")]
```
unleashed, line 63: Isn't this module already only compiled in when this is true?

cache-filter/src/filter/http.rs
```diff
@@ -1,9 +1,12 @@
+#[cfg(not(feature = "unique_callout"))]
+use crate::dummy::*;
```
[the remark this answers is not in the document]
rahulanand16nov: I was thinking of having a single dummy for `unqiue_callout` and `visible_logs` but `unique_callout_dummy` is also fine.

cache-filter/src/filter/root.rs
rahulanand16nov: Way better!

cache-filter/src/unique_callout.rs
rahulanand16nov: Forgot to remove it :weary:

cache-filter/src/filter/http.rs
rahulanand16nov: Workable!

cache-filter/src/lib.rs
rahulanand16nov: Done

rahulanand16nov:
Rebased and CI is passing locally.

unleashed (commented):
Ok @rahulanand16nov - now let's squash the reverts and the rename of the module with each diff belonging to each one of the relevant commits so the history is kept nice and clean!

cache-filter/Cargo.toml
```diff
@@ -16,6 +16,7 @@ crate-type = ["cdylib"]
 [features]
 default = ["prng_pcg32"]
 visible_logs = []
+unique_callout = []
```
unleashed, line 19: nitpick: typo in the commit message

rahulanand16nov: Fixed :D

rahulanand16nov:
@unleashed LGTM now, please approve and merge if it looks good to you as well :)

unleashed (approved):

unleashed:
Looks good!

PR #63, opened 2021-09-09
Benchmark report for high traffic scenario

NomadXD:
This PR includes benchmark reports for high traffic scenario of,

1. Proxy setup without unique callouts feature
2. Proxy setup with unique callouts feature


rahulanand16nov (approved):
Nothing much to review so looks good.

PR #65, opened 2021-09-15
Adding RateLimit headers and few simplifications

rahulanand16nov:
It was part of the TODOs in GSoC to add RateLimit headers, but it was put on hold to prioritize other work. I have tried to add a very basic implementation of them considering the complex 3scale data model. The filter will include these headers as long as there is no internal failure.

Let me know if I missed something. Thanks!

### DIFF
--- a/cache-filter/src/filter/http.rs
+++ b/cache-filter/src/filter/http.rs
@@ -219,10 +219,31 @@ impl HttpContext for CacheFilter {
         }
     }
 
-    #[cfg(feature = "visible_logs")]
     fn on_http_response_headers(&mut self, _: usize) -> Action {
-        let (key, val) = crate::log::visible_logs::get_logs_header_pair(self.context_id);
-        self.add_http_response_header(key.as_ref(), val.as_ref());
+        #[cfg(feature = "visible_logs")]
+        {
+            let (key, val) = crate::log::visible_logs::get_logs_header_pair(self.context_id);
+            self.add_http_response_header(key.as_ref(), val.as_ref());
+        }
+        // Adding RateLimit headers.
+        if self.state.rate_limit_info.limit.is_some() {
+            self.add_http_response_header(
+                "RateLimit-Limit",
+                &self.state.rate_limit_info.limit.unwrap().to_string(),
+            );
+        }
+        if self.state.rate_limit_info.remaining.is_some() {
+            self.add_http_response_header(
+                "RateLimit-Remaining",
+                &self.state.rate_limit_info.remaining.unwrap().to_string(),
+            );
+        }
+        if self.state.rate_limit_info.reset.is_some() {
+            self.add_http_response_header(
+                "RateLimit-Reset",
+                &self.state.rate_limit_info.reset.unwrap().to_string(),
+            );
+        }
         Action::Continue
     }
 }
--- a/cache-filter/src/filter/http.rs
+++ b/cache-filter/src/filter/http.rs
@@ -86,16 +86,26 @@ pub enum RequestDataError {
     UpstreamBuilderFail(#[from] threescalers::Error),
 }
 
+/// State data collected/updated during a request lifetime.
+#[derive(Clone)]
+pub struct RequestState {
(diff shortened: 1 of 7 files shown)


threescale/src/structs.rs
```diff
@@ -252,3 +252,24 @@ impl Message {
...
+    /// Header RateLimit-Reset's value.
+    pub reset: u64,
+}
+
+impl Default for RateLimitInfo {
+    fn default() -> Self {
+        RateLimitInfo {
+            limit: u64::MAX,
+            remaining: u64::MAX,
+            reset: u64::MAX,
```
unleashed, line 272: If any one of these is never set then it will have a maximum u64 value, which is not what we want. This is asking for you to use an Option and default it all to None, then emit headers when they are `Some(...)`.

threescale/src/utils.rs
```diff
@@ -30,7 +34,11 @@ pub fn limit_check_and_update_application(
                 let time_diff = current_time
                     .checked_sub(period.start)
                     .ok_or(UpdateMetricsError::DurationOverflow)?;
+
+                // This is atleast 1 because current time is higher than window end.
                 let num_windows = time_diff.as_secs() / period.window.as_secs();
+
+                // No. of secs to push forward window ends so the current time fall within new window.
```
unleashed, line 41: Add these comments in a separate commit.

cache-filter/src/filter/http.rs
```diff
@@ -58,7 +58,7 @@ enum AuthResponseError {
     ListKeysMiss,
     #[error("app id field is missing from list keys extension response")]
     ListAppIdMiss,
-    #[error("app id field is missing from list keys extension response")]
+    #[error("service id field is missing from list keys extension response")]
```
unleashed, line 61: change commit title to remove "make change to" - it is always implicit :)

cache-filter/src/filter/http.rs
```diff
@@ -267,8 +307,9 @@ impl CacheFilter {
                     self.resume_http_request();
                     break;
                 }
-                Err(UpdateMetricsError::RateLimited) => {
+                Ok(RateLimitStatus::RateLimited(rate_limit_info)) => {
                     info!(self.context_id, "request is rate-limited");
+                    self.state.rate_limit_info = rate_limit_info;
                     self.send_http_response(429, vec![], Some(b"Request rate-limited.\n"));
```
NomadXD, line 313: Here also we need to add rate limit headers. `send_http_response` uses local response which directly sends a response from here and destroy the filter object. `on_http_response_headers` gets triggered only  in the response path from upstream back to downstream.

rahulanand16nov: Actually Lahiru, this is not correct. I thought the same but after some testing, I saw `on_http_response_headers` is called even on the local response from `send_http_response`. You can even check the CI, in the rate-limit test, RateLimit headers are in fact sent back even though we are not adding it above. Interestingly, this fact allowed me to clean `visible-logs` up in the last PR as well ;)

cache-filter/src/filter/http.rs
rahulanand16nov: Oh Yea :sweat_smile:

threescale/src/structs.rs
rahulanand16nov: I was thinking Optioning the whole struct at first but thought about too many unwraps and checks but option for fields is better! Added.

threescale/src/utils.rs
rahulanand16nov: Splitted in action.

cache-filter/src/filter/http.rs
NomadXD: Ahh my bad !! It works because headers are added from the `on_http_response_headers` of the same filter. For filter chain `A -> B -> C`, if we do a local response from filter B, then the response path callbacks will get invoked from `B`  and downwards.

NomadXD (approved):

